PR #18, opened 2025-10-12
feat: add ClaudeCodeAdapter for Claude Code subscription billing

lakergreat1:
## 💰 Stop Burning Money on API Calls

**This PR enables Neo to use your Claude Code Max/Pro subscription instead of direct Anthropic API billing.**

If you're a Claude Code subscriber paying 0/month for unlimited usage, why are you ALSO paying -15 per million tokens for Neo? This adapter lets Neo tap into your existing subscription, potentially **saving hundreds of dollars per month** for heavy users.

**Real Cost Example:**
- Typical Neo query: ~18,000 tokens (context + response)
- Cost per query with API: ~$0.07 USD
- Cost per query with subscription: **$0.00**
- 100 queries/day savings: **~$210/month**

---

## Summary

Adds ClaudeCodeAdapter to enable Neo to use Claude Code Max/Pro subscriptions instead of direct Anthropic API billing.

## Implementation

- **ClaudeCodeAdapter class**: Shells out to `claude` CLI via subprocess
- **Subscription authentication**: Removes ANTHROPIC_API_KEY from environment to force OAuth
- **Streaming JSON parsing**: Parses `--output-format stream-json` output
- **Config validation**: Adds 'claude-code' to VALID_PROVIDERS list
- **Compatibility**: Accepts but ignores api_key/base_url parameters

## Benefits

- ✅ **Cost Savings**: Use Claude Code subscription instead of per-token API billing
- ✅ **Unified Billing**: Neo usage counts against same limit as web/mobile Claude
- ✅ **No API Key Management**: Subscription auth handled automatically by CLI
- ✅ **Proven Pattern**: Based on Cline's production implementation

## Technical Details

**Subprocess overhead**: ~100-500ms per call (negligible compared to 30-120s LLM inference)

**Based on**: Cline's implementation (https://github.com/clinebot/cline)

## Requirements

- Claude Code CLI installed: `npm install -g @anthropic-ai/claude-code`
- User authenticated: `claude auth login`
- ANTHROPIC_API_KEY should NOT be set (forces subscription auth)

## Usage

```bash
# Configure Neo to use Claude Code
neo --config set --config-key provider --config-value claude-code
neo --config set --config-key model --config-value claude-sonnet-4-5-20250929

# Use Neo normally (with subscription)
unset ANTHROPIC_API_KEY
neo "your query here"
```

## Testing

Tested locally with Claude Code Max subscription:
- ✅ No API billing charges
- ✅ Same response quality as direct API
- ✅ Works with semantic memory and pattern extraction
- ✅ Error handling for CLI issues

## Files Changed

- `src/neo/adapters.py`: Added ClaudeCodeAdapter class (~165 lines)
- `src/neo/cli.py`: Added 'claude-code' to VALID_PROVIDERS

## Backward Compatibility

✅ No breaking changes - adds new provider option without modifying existing adapters

## Reference

- Cline implementation: https://github.com/clinebot/cline/blob/main/src/integrations/claude-code/run.ts
- Claude Code CLI docs: https://docs.claude.com/en/docs/claude-code

### DIFF
--- a/src/neo/adapters.py
+++ b/src/neo/adapters.py
@@ -3,6 +3,10 @@ LM Adapter implementations for OpenAI, Anthropic, and local models.
 """
 
 import os
+import json
+import subprocess
+import tempfile
+from pathlib import Path
 from typing import Optional
 
 # Load environment variables from .env file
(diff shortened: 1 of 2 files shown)


mliotta:
Thanks for this, and apologies it sat open so long — that's on the repo, not on you.

Closing it as stale rather than as a judgement on the idea. Concretely, where it stands today:

- The head this PR references (`ce6032b`) is no longer reachable from the `feature/claude-code-adapter` branch, which now points at a `main` commit. GitHub still renders the original diff from the stored SHA, so the PR *looks* current when it isn't.
- It conflicts with `main` in 5 places. `src/neo/adapters.py` has changed substantially since October — the adapter layer picked up reactive model-parameter compatibility learning, CAR routing changes, and error-decoration work.
- There are no tests, and the adapter shells out to a CLI, which has meaningfully different failure modes from the HTTP adapters (process spawn, auth state, exit codes, timeouts) — the area most in need of them.

None of that is unfixable, but it adds up to a rewrite against today's `adapters.py` rather than a merge, and it's better to do that deliberately than to force-resolve conflicts in code without the author's intent to hand.

The underlying idea — routing inference through a Claude Code subscription instead of direct API billing — is still worth having. If you'd like to pick it back up against current `main`, please do open a fresh PR and I'll review it properly. Thanks again for taking the time.

PR #177, opened 2026-08-09
fix(index,engine): stop silently discarding most of the repo, and say so when a cap fires (#158, #159, #176)

### LINKED ISSUE #158: TypeScript/TSX `interfaces` query fails to compile — every interface is silently missing from the index
# TypeScript/TSX `interfaces` query fails to compile — every interface is silently missing from the index

**Evidence:** `neo --cwd ~/git/parslee-knowledge` (v0.41.0, 2026-07-30) ran to a 900s
`timeout` kill (exit 124) without returning an answer. Of 9,707 output lines, **9,699 were
one repeated warning**:

```
neo.index.language_parser WARNING: Failed to compile query typescript:interfaces: Invalid syntax
neo.index.language_parser WARNING: Failed to compile query tsx:interfaces: Invalid syntax
```

The run reached `[Neo] Invoking LLM inference...` and then spent the remaining ~13 minutes
in the indexing phase without producing output.

## Cause

`src/neo/index/language_parser.py:128` matches the TypeScript interface body as
`object_type`. The current `tree-sitter-typescript` grammar names that node
`interface_body`, so the query never compiles.

```
$ python -c "..."
node: interface_declaration | children: ['interface', 'type_identifier', 'interface_body']
FAIL object_type    -> Impossible pattern
OK   interface_body
```

`QUERIES['tsx'] = QUERIES['typescript']` (line 261) aliases the dict, so the single
definition breaks both languages.

## Reproducer

```python
from neo.index.language_parser import TreeSitterParser
from pathlib import Path

src = "export interface Foo { a: string; }\nexport class Bar { m() { return 1; } }\n"
chunks = TreeSitterParser().parse_file(Path("repro.ts"), src)
print([c.chunk_id for c in chunks])
```

Actual:

```
WARNING: Failed to compile query typescript:interfaces: Impossible pattern at row 2, column 17
['class:Bar']
```

Expected: `['class:Bar', 'interface:Foo']`

## Impact

The log flood and the stall are the visible symptom. The quieter problem is that
`interface Foo` is **dropped from the index without an error** — the parser returns the
class and moves on. Every Neo run over a TypeScript codebase since this shipped has been
reasoning about a tree with all of its interfaces invisible. For an interface-heavy repo
that is most of the type surface.

## Fix

```diff
  'interfaces': """
      (interface_declaration
          name: (type_identifier) @name
-         body: (object_type) @body) @interface
+         body: (interface_body) @body) @interface
  """
```

Verified locally: the warning disappears and both extensions extract correctly.

```
.ts  -> ['class:Bar', 'interface:Foo']
.tsx -> ['class:Bar', 'interface:Foo']
```

`tests/test_language_parser.py`, `tests/test_multilang_index.py` and
`tests/test_edge_extraction.py` — 45 tests — pass with the change.

End to end, on a repo with 175 `.ts`/`.tsx` files and 224 interface declarations:

| | before | after |
|---|---|---|
| outcome | exit 124, killed at the 900s cap | **exit 0 in 130s** |
| `language_parser` warnings | 9,699 | **0** |
| answer produced | none | confidence 0.95 |

Caveat on that comparison: the before-run was `--cwd` at a directory that nests ~574k files
and the after-run targeted a single repo, so the file-scan counts (59,759 → 537) reflect
scope, not the patch. The warning count and the completion do not — the after-run still
parsed 175 TypeScript files and emitted zero warnings where every one of them previously
failed.

## Why it went unnoticed

The line has been unchanged since `001e786` (2025-11-23, "feat: Add tree-sitter
multi-language code indexing" #63), and **no test references the `interfaces` query** —
`grep -rn "interfaces" tests/` returns nothing. The suite parses TypeScript fixtures but
only ever asserts on functions and classes, so a query that never compiles looks identical
to one that finds nothing.

Worth adding alongside the fix: assert that a TS fixture containing an interface yields an
`interface:` chunk. That is the assertion that would have caught this on day one, and it
would also catch the same class of breakage for `go:types`, `rust`, and the other
per-language queries when their grammars move.

## Environment

Versions are inside the declared pins in `pyproject.toml` — this is not resolution drift:

| | declared | installed |
|---|---|---|
| `tree-sitter` | `>=0.23,<0.26` | 0.25.2 |
| `tree-sitter-language-pack` | `>=0.13.0,<1.0` | 0.13.0 (the floor) |

macOS 26.5.2 (darwin 25.5.0), Python 3.14.0, neo-reasoner 0.41.0 via pipx from PyPI.

Since the installed language-pack is at the *minimum* allowed version, a fresh install
today resolves to the same or newer grammar and fails identically.

### LINKED ISSUE #159: `neo --index` on a .NET repo produces an index containing zero C#
# `neo --index` on a .NET repo produces an index containing zero C#

**Evidence:** two `neo --index` runs on 2026-07-30, both exit 0, no errors reported to the
user. The resulting indexes:

| repo | files indexed | composition | repo actually contains |
|---|---|---|---|
| m365dotnet | 83 | **100% `.py`** | 4,272 `.cs`, 54 `.ts/.tsx`, 18 `.py` |
| odi | 92 | 28 `.py`, 64 `.cs`, **0 `.ts`** | 1,383 `.cs`, 175 `.ts/.tsx` |

m365dotnet is a .NET backend. Its semantic index contains **no C# at all**. It has only 18
tracked Python files, so the 83 indexed came from elsewhere:

```
.worktrees/PAR-7426-u7/adws/poll_prs.py
.worktrees/PAR-7426-u7/adws/review_pr.py
.worktrees/PAR-7426-u7/scripts/copy-cosmos-data.py
```

A git worktree checked out inside the repo. ODI shows the same shape — its index is
largely `.claude/dorothy-review/corpus/tools/*.py` plus the *same files again* from
`.claude/worktrees/issue-393-.../`.

Both runs printed `✔ Built index` and exited 0. Nothing signals that the index does not
represent the repository.

## Two causes, both still on `main`

**1. Language-ordered concatenation, truncated without ranking**
(`ProjectIndex.build_index`, `src/neo/index/project_index.py:337-343`)

```python
files_to_index = []
for pattern in file_patterns:
    files_to_index.extend(self.repo_root.glob(pattern))

files_to_index = files_to_index[:max_files]
```

`file_patterns` is a fixed list beginning `**/*.py`, `**/*.cs`, `**/*.ts`. Every Python
file in the tree is appended before the first C# file, so the cut keeps whichever
languages happen to glob first. On m365dotnet, Python from a worktree exhausted the budget
and `**/*.cs` was never reached. On ODI, Python plus C# exhausted it and `**/*.ts` was
never reached — which is why 175 TypeScript files produced zero indexed TypeScript.

The cut is also unranked: no centrality, no size, no recency, no path heuristic. It is
whatever `Path.glob` yields first.

**2. No path exclusions at all**

`grep -n "worktree\|\.claude\|node_modules\|exclude" src/neo/index/project_index.py`
returns nothing. Git worktrees, `.claude/` tooling corpora, vendored trees and build
output are all eligible. Worse, a worktree is a *second copy of the same repository*, so
identical files consume slots twice — visible in the ODI sample above, where
`cohens_kappa.py` appears both at its real path and under `.claude/worktrees/...`.

## Suggested fix

Rank before truncating instead of relying on glob order, and exclude non-source paths:

- Interleave or score across languages so the cut cannot eliminate an entire language.
  Weighting by repo composition would put ~4,272 C# files ahead of 18 Python ones in a
  .NET repo without any language-specific special-casing.
- Skip `.worktrees/`, `.git/`, `node_modules/`, `bin/`, `obj/`, `dist/`, `.claude/`, and
  de-duplicate by content hash so a worktree copy cannot occupy a second slot.
- When the cap truncates, say so: `Built index: 83 files (capped at 100; 4,326 eligible —
  0 of 4,272 .cs files included)`. Silent truncation is what makes this look like success.

## Note on `--max-files`

Not reporting this one — `main` already has
`max_files = getattr(args, 'max_files', None) or 100`, so the flag is wired. The released
0.41.0 on PyPI still carries `max_files = 100  # Configurable later`, so the fix is
unshipped rather than missing.

Raising the cap does not resolve the above on its own: ordering still front-loads one
language, and worktree duplicates still consume the larger budget.

## Impact

`--index` is advertised in normal runs (`[Neo] Tip: run 'neo --index' to enable semantic
file selection`), so users are directed toward it. Taking that advice on either repo above
produces an index that cannot answer a question about the primary language.

The index is only consulted under `--semantic`
(`src/neo/cli.py:927`), which limits the blast radius — but it inverts the value of the
flag. Passing `--semantic` on m365dotnet would restrict Neo to 83 Python files from a stale
worktree instead of keyword-scoring the C# it is being asked about, making results worse
than not indexing at all.

## Environment

neo-reasoner 0.41.0 (pipx, PyPI), macOS 26.5.2, Python 3.14.0. Line references are against
`main` @ `140929b`; the PyPI build's line numbers differ.

Chunk cap `MAX_CHUNKS_PER_REPO = 1000` (`project_index.py:52`) also fired on m365dotnet
(`Too many chunks (1163), limiting to 1000`) — noted for context, not part of this report.

### LINKED ISSUE #176: Context files are silently truncated to 3000 chars with no marker, and the banner reports pre-truncation bytes — the model reasons about a fragment believing it has the file
`src/neo/engine.py:1614-1620` truncates every context file to its first 3000 characters, appends no marker, and reports the pre-truncation byte count to the operator. The model therefore reasons about the top of a file while believing it has the whole thing, and the console says it was given more than it was.

```python
parts.append(f"\nREPOSITORY CONTEXT ({len(files)} files, {sum(len(f.content or '') for f in files)} bytes):")
for f in files[:20]:
    is_important = any(pat in f.path.lower() for pat in ['readme.md', 'claude.md', 'architecture'])
    char_limit = 8000 if is_important else 3000
    content_preview = (f.content or '')[:char_limit]
    parts.append(f"\n--- {f.path} ---\n{content_preview}")
```

## Three separate problems

**1. The truncation is unmarked.** `content_preview` is a bare slice. The model receives `--- path ---` followed by text that simply stops mid-token, with nothing indicating the content was cut. It cannot distinguish "this file ends here" from "this file was clipped," so it will reason about absence — "X is not called anywhere in this file," "no null check is present" — from a fragment.

`neo/agent_context.py:162-163` already does this correctly for instruction docs:

```python
if len(content) > PER_FILE_CAP_BYTES:
    content = content[:PER_FILE_CAP_BYTES] + "\n... [truncated]"
```

The source-file path should do the same.

**2. The byte count is computed before the cut.** The banner sums `len(f.content)` at line 1614; the truncation happens at 1619. An operator reading `REPOSITORY CONTEXT (12 files, 340000 bytes)` concludes the model saw 340 KB. With a 3000-char cap and a 20-file limit, the model saw at most 60 KB, and usually far less. There is no way to tell from the output that the number is not what was sent.

**3. 3000 characters is roughly 40-75 lines of C#.** Anything below that in a file is invisible. In practice that means constants, `using` blocks, type declarations and the first method or two are visible; everything else is not.

## How this surfaces

Running claim-verification against a C# codebase (files 15-100 KB), the failure was consistent and initially looked like a retrieval-tuning problem:

- Claims about constants and signatures near the top of a file: answered correctly, first pass, high confidence.
- Claims about method bodies past ~line 60: neo repeatedly and correctly refused, reporting the body was not in its context rather than guessing from XML doc comments. That refusal behavior is good and worth preserving.
- One claim landed with its cutoff on a method *signature*, body out of view. Neo confirmed it at 0.90 — plausible, and the wording turned out to be wrong. Supplying the body moved it to 0.99 and produced a correction. **This is the dangerous case: enough context to look answerable, not enough to be right.**

None of `--max-files`, `--max-bytes`, `--include`, or prompt keyword tuning affects this, because all of them act upstream of line 1619. `context_gatherer.py` may select a well-scored 200-line window and have it clipped to its first 40 lines here regardless.

## Suggested fix

Minimal, in priority order:

1. **Append a marker when truncating**, matching `agent_context.py`:
   ```python
   content = f.content or ''
   if len(content) > char_limit:
       content = content[:char_limit] + f"\n... [truncated: {len(content) - char_limit} more chars]"
   ```
2. **Report post-truncation bytes in the banner**, or report both (`sent X of Y bytes`). Also worth surfacing when `files[:20]` drops files.
3. **Consider raising the cap or making it configurable.** 3000 is low for compiled-language source. If a total-budget ceiling is the real constraint, a budget divided across selected files would beat a flat per-file cap — a run that selects 3 files could afford far more per file than one selecting 20.
4. **Consider `is_important` by extension rather than filename substring.** A path containing "architecture" gets 8000 chars whether or not it is a doc; a 40 KB service class gets 3000.

Item 1 alone would have converted every silent-fragment answer in the session above into an honest refusal, which neo already handles well.

## What changes when this lands

- A truncated file carries a visible marker, and the model can say "I cannot see the rest of this file" instead of inferring from a fragment.
- The banner's byte count matches what was actually sent.
- A test asserts that a file exceeding `char_limit` produces a prompt containing the marker, and that the reported byte count equals the sum of what was included.

Found while using `neo --mode advise` to verify ~50 claims about C# implementation details across several repos. Environment: neo 0.41.0 (pyproject) / installed venv at `~/.local/pipx/venvs/neo-reasoner`, Python 3.14, macOS.

JLiotta-25:
## Description

Fixes three independent defects that share one shape: **the index and the prompt both silently discarded most of what they were asked to represent, and reported success.**

- **#158** — four tree-sitter queries did not compile against the installed grammars. A query that fails to compile is indistinguishable from one that matches nothing, so TypeScript/TSX interfaces and *every* C# inheritance edge had been missing from every index since the queries shipped. The only signal was a warning emitted once per query **per parsed file** — 9,699 lines in one reported run.
- **#159** — `build_index` globbed file patterns in list order, concatenated, and sliced to `--max-files`. Whichever language globbed first ate the budget: a .NET repo of 4,272 C# files produced an index of 83 Python files (95 of them from a git worktree checked out inside the repo) and zero C#, exit 0.
- **#176** — every context file was clipped to 3000 characters with no marker, while the banner reported the **pre-truncation** byte count. The model reasoned about the top of a file believing it had the whole thing.

## Type of Change

- [x] Documentation update

## Related Issue

Fixes #158
Fixes #159
Fixes #176

### #158 — queries that never compiled

`typescript`/`tsx` `interfaces` matched the interface body as `object_type`; the grammar calls it `interface_body`. Probing every entry in `QUERIES` and `EDGE_QUERIES` found three more broken the same way:

| query | fault | consequence |
|---|---|---|
| `typescript`/`tsx` `interfaces` | `object_type` → `interface_body` | every TS interface missing from the index |
| `c_sharp` `inheritance` | `bases:` is not a field in the grammar | every C# inheritance edge missing |
| `python` `module_doc` | matched an `expression_statement` wrapper the grammar does not produce | none — see below |

`module_doc` was **deleted rather than fixed**: correcting it only revealed that nothing consumes a lone `@doc` capture (`_process_captures` builds chunks from construct captures), so it has never produced a chunk in any released version.

Compile results are now cached **including failures**, which is what turns 9,699 warnings into one per query per process.

C# bases additionally needed widening beyond the reported fault — an identifier-only, class-only pattern compiles fine and silently drops `: Repository<Order>` (parses as `generic_name`) and `interface IX : IY` entirely. Now matches `class`/`interface`/`record`/`struct` with generic bases.

### #159 — a slice is not a ranking

`_select_files` now groups eligible files by language and gives each a share of `--max-files` proportional to repo composition, with a floor of one slot per language (`_allocate_slots`). Non-source paths are excluded on two layers: `EXCLUDED_DIR_NAMES` for what repos forget to ignore (`.worktrees`, `.claude`, `node_modules`, `obj`, virtualenvs), plus the repo's own `.gitignore` via the helper `context_gatherer` already uses. Byte-identical files are indexed once, and the freed slot refills — across languages if the original has nothing unique left.

Ambiguous directory names — `bin`, `build`, `out`, `target`, `dist`, `vendor` — are deliberately **not** hardcoded. Each is real source somewhere (`src/bin/main.rs`, vendored trees); a sweep of three local repos found 254 git-tracked source files under `bin/` or `vendor/` alone. Repos that generate into them gitignore them, and the asymmetry is one-sided: over-excluding hides code permanently and invisibly, over-including only spends slots.

**Fixing the file cut alone was not enough.** `MAX_CHUNKS_PER_REPO` re-created the identical bug one function later: chunks arrive grouped by file and files grouped by language, so `chunks[:1000]` kept 1000 C# chunks and discarded every TypeScript and Python chunk the allocator had just fought to include. `_cap_chunks` round-robins across files instead.

Measured on a synthetic repo shaped like the reported one — 300 C#, 20 TS, 5 Python, plus a 200-file Python worktree under `.worktrees/`, `--max-files=100`:

| | before | after |
|---|---|---|
| files selected | 100, all `.py`, 95 of them from the worktree | `.cs` 91 / `.ts` 7 / `.py` 2, worktree excluded |
| chunks kept after the 1000 cap | `.cs` 1000, `.ts` 0, `.py` 0 | `.cs` 984 / `.ts` 14 / `.py` 2 |
| selected files represented in the index | 63 of 100 | 100 of 100 |

The per-repo figures in the issue reports (4,272 `.cs`; 83 indexed; 175 `.ts`) are the reporter's, not re-measured here.

### #176 — an unmarked cut invites a wrong inference

Every truncation now carries `... [truncated: N of M characters not shown]`, matching what `agent_context._read_doc` has always done. The banner reports post-truncation size, and says **chars** rather than bytes — `len()` on a `str` counts code points, so the old label was wrong twice over.

`_render_context_files` returns a third value, `visible`: each shown file cut to what the model actually received. `code_smells.scan_files` is fed that instead of the originals, because scanning full content emitted findings like `src/Service.cs:401 [todo/warn] HACK: ...` for a file the model was shown to line ~215. An unmarked cut *invites* a wrong inference about unseen code; a line-numbered finding about unseen code *asserts* one.

## How Has This Been Tested?

- [x] `1905 passed, 8 skipped` — full suite, against a measured `1750 passed, 8 skipped` baseline on `main` (+155 tests)
- [x] `ruff check src/ tests/ tools/` clean
- [x] Every new regression test verified to **fail** against the pre-fix source, so none is vacuous — including reverting each fix in turn to confirm the guarding test goes red
- [x] Both original reporter scenarios reproduced end to end and confirmed fixed (tables above)
- [x] Reviewed by the `neo` CLI itself and by an adversarial reviewer; both re-verified after fixes

New test files: `tests/test_index_file_selection.py` (64 tests), `tests/test_context_file_truncation.py` (24 tests).

`test_every_chunk_query_compiles` / `test_every_edge_query_compiles` are parametrized over **every** entry in `QUERIES` and `EDGE_QUERIES`, so the next grammar rename fails a test instead of quietly shrinking the index. They prove compilation only, so behavioural assertions were added alongside for the constructs that actually broke.

**Test Configuration**:
- Python version: 3.13.9
- OS: macOS 26.5.2 (darwin)
- Neo version: 0.43.0

## Checklist

## Additional Notes

**Behaviour changes an operator will notice:**

1. `neo --index` now prints what the caps left out — eligible vs indexed with a per-language breakdown, excluded paths, skipped duplicates, a capped chunk total, and files that contributed no chunks. Every line is conditional, so a build that indexed everything prints none of them and their presence is the signal. "Built index" on its own no longer means the index represents the repository.
2. Indexes built before this change should be rebuilt (`neo --index`); an existing `.neo/index.json` still carries the old composition.
3. `docs/tree-sitter-setup.md` claimed "Neo honors `.gitignore` by default" for the index. That was false — the handling existed only in `context_gatherer`. It is now true, via the same helper.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -559,6 +559,22 @@
      files; `MAX_CHUNKS_PER_REPO` is fixed at 1000 while `--max-files` is not,
      so `files_with_chunks` reports the shortfall — `truncated` is False there,
      because the FILE cap genuinely was not the constraint.
+     **And the corollary the first version of this report broke: never blame a
+     cap for an absence it did not cause.** A selected file is missing from the
+     index for one of TWO unrelated reasons, and the console must not guess
+     between them. Either it produced no chunks at all — no function, class,
+     interface or struct for the grammar to match, as in an empty
+     `__init__.py`, an enum-only `.cs`, a type-alias-only `.ts` — which no cap
+     setting changes and which therefore gets a bare statement with NO remedy
+     attached; or the cap took every chunk it produced, which gets the cap
+     named and `lower --max-files`. `files_producing_chunks` is measured
+     BEFORE `_cap_chunks` precisely so the two stay separable; subtracting the
+     post-cap `files_with_chunks` from `selected` conflates them and is what
+     printed "the 1000-chunk cap is below the 25 files selected" for a build
+     that kept 559 of 559 chunks, with `chunks_capped` False on the same
+     report. It fired on this repo, on any repo with an `__init__.py`. A report
+     that invents a cause is worse than the silence it replaced, because
+     silence at least does not send the operator to the wrong knob.
   **Query footgun** (`language_parser.py`; incident detail in the tree-sitter
   doc's "Why queries break silently"): a query that fails to compile is
   indistinguishable from one that matches nothing — `_get_query` returns None
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -520,6 +520,58 @@
   and retrieval changes than waiting for an inference round trip. **Use it before
   believing any claim about what Neo "saw"** — the two defects below were both
   invisible from the outside and presented as the model being unhelpful.
+- Project index (`index/project_index.py`, `index/language_parser.py`; full
+  notes in `docs/tree-sitter-setup.md`). Three invariants, each of which was
+  violated and each of which produced an index that could not answer a question
+  about its own repository:
+  1. **Budgets are apportioned, never sliced.** `_select_files` groups eligible
+     files by language and hands each a share of `--max-files` proportional to
+     repo composition, floor of one slot per language (`_allocate_slots`); then
+     `_cap_chunks` round-robins `MAX_CHUNKS_PER_REPO` across FILES so each keeps
+     a chunk before any keeps a second. Both exist because a list slice is not a
+     ranking: globbing `**/*.py` before `**/*.cs` and slicing gave a .NET repo of
+     4,272 C# files an index of 83 Python files (95 from an in-repo worktree) and
+     zero C#, exit 0. Fixing only the file cut left `chunks[:1000]` re-creating it
+     one function later — chunks arrive grouped by file and files by language, so
+     the slice kept 1000 C# chunks and dropped every other language, with 37 of
+     the 100 selected files contributing nothing.
+  2. **Exclusion is two layers and `bin`/`build`/`out`/`target`/`dist`/`vendor`
+     belong to neither by name.** `EXCLUDED_DIR_NAMES` covers what repos forget to
+     ignore (`.worktrees`, `.claude`, `node_modules`, `obj`, virtualenvs);
+     `_build_exclusion_filter` layers the repo's own `.gitignore` on top via
+     `context_gatherer.load_gitignore_patterns`. **Footgun**: `should_ignore` only
+     tests the path handed to it, so ancestor directories must be walked
+     separately — `iter_paths` gets this via `os.walk` pruning, which the index
+     has no equivalent of. Matching is exact and case-sensitive, against
+     directory components only. The ambiguous names stay out because each is real
+     source somewhere (`src/bin/main.rs`, vendored trees; 254 tracked files under
+     `bin/`+`vendor/` across three local repos) and the asymmetry is one-sided:
+     over-excluding hides code permanently, over-including only spends slots.
+  3. **A cap that fired must be reported.** `selection_report` carries eligible /
+     selected / excluded / duplicates / chunk counts and the CLI prints them.
+     `truncated` means THE CAP BOUND US and is `examined < eligible` — whether
+     any candidate went unlooked-at, which the per-language iterators already
+     know. Both cheaper predicates name the wrong knob: `selected < eligible`
+     made dedup print "2 of 7 eligible files (capped at --max-files=1000)", and
+     `selected >= max_files` did the same whenever `--max-files` landed exactly
+     on the unique-file count. Raising a cap that never bound fixes nothing.
+     Separately, round-robin only represents every file while chunk slots ≥
+     files; `MAX_CHUNKS_PER_REPO` is fixed at 1000 while `--max-files` is not,
+     so `files_with_chunks` reports the shortfall — `truncated` is False there,
+     because the FILE cap genuinely was not the constraint.
+  **Query footgun** (`language_parser.py`; incident detail in the tree-sitter
+  doc's "Why queries break silently"): a query that fails to compile is
+  indistinguishable from one that matches nothing — `_get_query` returns None
+  and `parse_file` moves on; edge failures log at DEBUG. Four were broken at
+  once, so TS interfaces and ALL C# inheritance edges were absent from every
+  index since they shipped. Three rules follow. Compile results are cached
+  INCLUDING failures (the uncached retry warned per query *per file* — 9,699
+  lines in one run). C# bases need `[(identifier) (generic_name (identifier))]`
+  across class/interface/record/struct, since an identifier-only, class-only
+  pattern compiles fine and silently drops `: Repository<Order>` and
+  `interface IX : IY`. And `test_every_chunk_query_compiles` /
+  `test_every_edge_query_compiles` prove compilation ONLY, so a new query still
+  needs its own behavioural assertion.
 - Context selection (`context_gatherer`): **a path named in the prompt is pinned**
   (`EXPLICIT_PATH_BOOST=10.0`, chosen to exceed every organic signal combined —
   filename overlap caps at +1.8, the re-rank boosts at +1.0 and +1.2). Without it a
@@ -553,6 +605,33 @@
   global budget is DROPPED — the original bug returning through a different door.
   `_fit_to_budget` shrinks outward from the window's best line so truncation can never
   discard the line that earned the window.
+- Prompt-side file rendering (`engine._render_context_files`): the REPOSITORY
+  CONTEXT block caps each file at `_CONTEXT_FILE_CHARS` (3000), or
+  `_IMPORTANT_FILE_CHARS` (8000) when the path contains one of
+  `_IMPORTANT_FILE_PATTERNS`, and shows at most `_MAX_CONTEXT_FILES` (20).
+  **Every cut MUST be marked and the banner MUST count post-truncation
+  characters** — that is the whole contract, and both halves were broken.
+  `content_preview = content[:char_limit]` appended nothing, so `--- path ---`
+  followed by text that just stops was indistinguishable from a file that ends
+  there, and the model answered questions about absence ("X is never called",
+  "there is no null check") from a fragment. The dangerous case is not an
+  obvious clip but a cut landing just past a method signature: enough to look
+  answerable, not enough to be right — measured live, a claim confirmed at 0.90
+  that moved to 0.99 *and reversed* once the body was supplied. Marked
+  truncation converts that into a refusal, which neo already does well.
+  Meanwhile the banner summed `len(f.content)` BEFORE the cut and called the
+  result bytes, so "12 files, 340000 bytes" described a payload the model never
+  received. It now reads `sent of total chars` — **chars, not bytes**: `len()`
+  on a `str` counts code points and the caps are character counts, so claiming
+  bytes was a second lie on any non-ASCII source. **Footgun**: the function
+  returns THREE values — `(sections, banner, visible)`. `visible` is each shown
+  file cut to what the model actually got, and `code_smells.scan_files` MUST be
+  fed it rather than the originals. Scanning full content emitted findings like
+  `src/Service.cs:401 [todo/warn] HACK: ...` for a file the model saw to line
+  ~215: an unmarked cut invites a wrong inference about unseen code, but a
+  line-numbered finding about unseen code asserts one. `neo.agent_context`
+  marks its truncation (`PER_FILE_CAP_BYTES` + `... [truncated]`) — copy that
+  shape for any new path that cuts text. Use `neo --dry-run` to see the block.
 - Host communication (`neo.events` + `models.OrchestratorMessage`): the engine
   reports lifecycle facts so a host (Claude Code, MCP, an IDE) doesn't have to
   reverse-engineer presentation from raw plans. **Governing rule: Neo reports
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -582,10 +582,15 @@
   once, so TS interfaces and ALL C# inheritance edges were absent from every
   index since they shipped. Three rules follow. Compile results are cached
   INCLUDING failures (the uncached retry warned per query *per file* — 9,699
-  lines in one run). C# bases need `[(identifier) (generic_name (identifier))]`
-  across class/interface/record/struct, since an identifier-only, class-only
-  pattern compiles fine and silently drops `: Repository<Order>` and
-  `interface IX : IY`. And `test_every_chunk_query_compiles` /
+  lines in one run). C# bases need all four of `identifier`, `generic_name`,
+  and `qualified_name` with either as its `name:` field — across
+  class/interface/record/struct — since a narrower pattern compiles fine and
+  silently drops `: Repository<Order>`, `: System.Exception` and
+  `interface IX : IY`. That query is GENERATED from `_CS_BASE_TYPES` over the
+  four declaration kinds rather than written out four times: the hand-copied
+  version is what left `interface_declaration` uncovered, and each widening
+  since has had to be applied everywhere at once or not at all. And
+  `test_every_chunk_query_compiles` /
   `test_every_edge_query_compiles` prove compilation ONLY, so a new query still
   needs its own behavioural assertion.
 - Context selection (`context_gatherer`): **a path named in the prompt is pinned**
--- a/README.md
+++ b/README.md
@@ -659,7 +659,7 @@ Neo is one binary with a plain-text prompt plus a handful of subcommands. Run
 | `--index` / `--update` / `--languages CSV` | Build, incrementally refresh, and scope the per-project semantic index |
 | `--semantic` | Use the semantic index for file selection (requires `.neo/index.json`) |
 | `--max-bytes N` | Hard cap on total context bytes (default 300000) |
-| `--max-files N` | Cap on files: context gathering (default 30), or the index build when passed with `--index` (default 100) |
+| `--max-files N` | Cap on files: context gathering (default 30), or the index build when passed with `--index` (default 100). The index build apportions this budget across languages by repo composition and reports what the cap left out |
 | `--include GLOB` / `--exclude GLOB` | Allow/block file patterns; both repeatable |
 | `--exts CSV` | Restrict context to these file extensions |
 | `--diff-since REV` | Prioritize files changed since a git rev or duration |
@@ -1130,6 +1130,8 @@ The context gatherer picks files using three signals:
 - **EPISODE-history feedback loop**: each Neo run stashes touched file paths as `file:<rel>` tags on EPISODE facts. On the next run, the gatherer queries for similar past prompts and gives those files up to +0.5 boost — past file selections measurably influence future ones.
 
 A per-file chunk cap of 2 prevents large files from eating the budget; a one-time first-run hint fires if the index is missing.
+
+The index build itself apportions its budget rather than truncating whatever globbed first: files are grouped by language and given a share of `--max-files` proportional to how much of the repo they are, with a floor of one slot per language, and the `MAX_CHUNKS_PER_REPO` cut round-robins across files so every indexed file keeps a chunk. Non-source paths (`.worktrees`, `node_modules`, `.claude`, virtualenvs, plus anything the repo's own `.gitignore` names) are excluded, and byte-identical duplicates are indexed once. When a cap bites, `neo --index` says so — a capped build no longer looks the same as a complete one. See [tree-sitter setup](docs/tree-sitter-setup.md#operational-notes).
 
 
 ### Learning Feedback Loop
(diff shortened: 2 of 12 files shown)


mliotta (changes requested):
Reviewed by checking the branch out and running it, not just reading the diff. CI is green on 3.10–3.13 and all three fixes are real — I confirmed each mechanism holds. **One defect needs fixing before merge, and it is in the new code.**

## Blocking: the new operator report asserts a false cause on almost any repo

`_format_selection_report` emits the shortfall line whenever `selected > files_with_chunks`:

```python
if 'files_with_chunks' in report:
    missing = report['selected'] - report['files_with_chunks']
    if missing > 0:
        lines.append(
            f"[Neo] {missing} selected files contributed no chunks — the "
            f"{report['max_chunks']}-chunk cap is below the "
            f"{report['selected']} files selected; lower --max-files"
        )
```

The comment above it says this is "only possible when the file count exceeds the chunk cap." That is not true. `_extract_chunks_from_file` returns `[]` for any file with no extractable constructs — `__init__.py`, an enum-only `.cs`, a type-alias-only `.ts`, an unparseable file — entirely independent of any cap.

Reproduced on this repository, on this branch, with the chunk cap nowhere near binding:

```
{'selected': 25, 'eligible': 178, 'chunks_extracted': 559, 'chunks_kept': 559,
 'chunks_capped': False, 'max_chunks': 1000, 'files_with_chunks': 24}

[Neo] Indexed 25 of 178 eligible files (capped at --max-files=25)
[Neo]   python 25/178
[Neo] 1 selected files contributed no chunks — the 1000-chunk cap is below the
      25 files selected; lower --max-files
```

1000 chunk slots against 25 files. The cap did not fire — `chunks_capped` is `False` on the same report. The culprit was `tests/__init__.py`, which is empty. It fires at the default `--max-files=100` too (3 files there).

Two things are wrong, and the second is worse than the first:

1. The line names a cause that was demonstrably not operating, on evidence the report itself contradicts one key over.
2. The remedy it prescribes is backwards. Lowering `--max-files` indexes strictly less and cannot make an empty `__init__.py` produce a chunk.

This is the failure class the PR exists to eliminate — output that reports confidently and wrongly — reintroduced in the reporting layer added to fix it. It will fire on essentially every Python repo.

The four tests around this line don't catch it because every fixture file in them is written as `def f{i}(): return {i}`, so no fixture is ever construct-less. `test_files_left_without_chunks_are_reported` patches `MAX_CHUNKS_PER_REPO` down to 2 to force the cap case, which is the right test — it just leaves the other cause uncovered.

Suggested shape: attribute the cap only when it actually bound (`chunks_capped and max_chunks < selected`), and give the ordinary case its own line — "N selected files yielded no extractable constructs" — with no remedy attached, since there isn't one.

## Non-blocking

**C# `qualified_name` bases still produce no edge.** The widened alternation covers `identifier` and `generic_name` but not `qualified_name`, so a fully-qualified base is dropped. Verified on this branch:

```
class A : System.Exception   -> no edge
class B : Repository<Order>  -> B inherits Repository   ✓
class C : IFoo               -> C inherits IFoo         ✓
interface IX : IY            -> IX inherits IY          ✓
class D(int x) : BaseThing(x)-> D inherits BaseThing    ✓
```

`: System.Exception` and `: Microsoft.AspNetCore.Mvc.ControllerBase` are everyday .NET. This is strictly better than `main`, which produced zero C# inheritance edges of any kind, so it's a follow-up rather than a blocker — but it's the same shape as the fault the commit message calls out for `generic_name`, and worth closing while the reasoning is fresh.

**`glob` still walks the excluded directories.** `EXCLUDED_DIR_NAMES` filters after `repo_root.glob('**/*.cs')` has already materialized every match, so a repo with a large `node_modules` or in-repo worktree still pays the full traversal — the exclusion buys correct *selection*, not the walk cost. Not a regression, and `iter_paths`-style pruning is a bigger change than this PR should carry; noting it so the next person doesn't assume the traversal is bounded.

**`replace(f, content=...)` leaves `bytes` stale** on the `visible` copies of `context_gatherer.ContextFile`. Harmless today — `scan_files` reads only `path` and `content` — but `visible` is now a value other code can reach for, and a `ContextFile` whose `bytes` disagrees with its `content` is a trap for whoever does.

## What's right

`_allocate_slots` is correct on the edges I traced: it can't over-allocate (floors sum to at most `shared`), surplus redistributes, and the refill loop terminates. `truncated = examined < eligible` is the right predicate, and the comment explaining why both cheaper alternatives lie is accurate rather than decorative — I checked the all-duplicates case (`examined == eligible`, `truncated` correctly `False`, duplicates reported separately).

Caching compile *failures* is the actual fix for the 9,699-line flood; caching only successes would not have been. The `_get_parser` locals-then-publish change fixes a real KeyError-masking-the-real-error bug that wasn't in any of the three issues.

`_render_context_files` verified end to end: marker present and accurate, banner counts post-truncation characters, `code_smells` fed the truncated text. Feeding `scan_files` the visible content rather than the originals is the detail most reviewers would have skipped, and it's the one that stops the tool asserting line numbers the model never saw.

mliotta:
Pushed two commits to this branch rather than leaving the review to bounce.

**`6003e91` — the blocking finding.** `files_producing_chunks` is now recorded *before* `_cap_chunks`, which is what makes the two causes separable at all. Cap starvation keeps the cap's name and the `lower --max-files` remedy; a constructs-free file gets a bare statement with no remedy attached, because there isn't one. The original reproduction now reads:

```
--- max_files=25 ---
[Neo] Indexed 25 of 178 eligible files (capped at --max-files=25)
[Neo]   python 25/178
[Neo] 1 selected file yielded no chunks (no functions, classes or interfaces for the parser to extract)

--- max_files=100 ---
[Neo] Indexed 100 of 178 eligible files (capped at --max-files=100)
[Neo]   python 100/178
[Neo] Kept 1000 of 2413 chunks (cap 1000); each indexed file keeps its first chunks before any file keeps a second
[Neo] 3 selected files yielded no chunks (no functions, classes or interfaces for the parser to extract)
```

The `max_files=100` case is worth reading closely: the chunk cap *did* fire, and the report says so — but round-robin still gave all 97 chunk-producing files a chunk, so no file was starved and the cap is correctly not blamed for the 3 that had nothing to give.

Four new tests, including the two the existing fixtures couldn't reach: a constructs-free file with no cap firing, and both causes present at once. Plus one pinning that a report predating the new key produces silence rather than a fabricated cap event — the fallback resolves `starved` to zero, which is the safe direction.

**`46a9e36` — the `qualified_name` follow-up**, since it's the same fault one node type over and the reasoning was already loaded:

```
class A : System.Exception      -> nothing        now A -> Exception
class B : My.Ns.Repo<Order>     -> nothing        now B -> Repo
class C : global::Foo.Bar       -> nothing        now C -> Bar
```

The capture is the rightmost segment via the `name:` field, so qualified and unqualified bases land in the graph under one naming convention rather than two. The query is now generated from `_CS_BASE_TYPES` over the four declaration kinds instead of written out four times — the hand-copied version is what left `interface_declaration` uncovered originally, and this commit would otherwise have had to widen four places by hand, which is the same trap. Also pinned the negative: a declaration with no base list yields no edge.

`1912 passed, 8 skipped`, ruff clean against the pinned 0.16.1.

Left alone, as noted in the review: the `glob`-walks-excluded-directories cost (real, but `iter_paths`-style pruning is bigger than this PR should carry) and the stale `bytes` field on `visible` copies.

---

One thing to know before you touch this branch: **committing from a linked worktree currently corrupts the repository** — #179, fixed on `main` in #180 but not present here. It bit while I was preparing these commits; `git commit` in a worktree ran the pre-commit suite with an inherited `GIT_DIR`, and `tests/test_pending_session_retention.py`'s `git init` reinitialized the real gitdir, set `core.bare = true` and wrote fixture commits onto the branch. Both commits above were made from the main checkout instead, which does not export `GIT_DIR` to hooks. Merging `main` in after #180 lands will clear it.

mliotta (approved):
Blocker resolved — I pushed the fix to this branch in `6003e91` and the C# `qualified_name` follow-up in `46a9e36`. Green on all four Python versions. Approving and merging; the two deferred items (glob traversal cost, stale `bytes` on `visible` copies) are noted in the review and not worth holding this.

PR #178, opened 2026-08-09
fix(prompts): mark every truncation that reaches a model

JLiotta-25:
## Description

> **Stacked on #177.** Base is `fix/index-selection-and-context-truncation`, because the `engine.py` regions are adjacent. GitHub retargets this to `main` when #177 merges.

#176 fixed one unmarked truncation. Auditing the codebase for the same defect class found **seventeen more cuts across eight prompt builders in six modules** — every one of them handing a model text that simply stops, indistinguishable from text that ended.

The model then reasons about absence from a fragment. The dangerous case is not an obvious clip but a cut landing just past a signature: enough to look answerable, not enough to be right.

## Type of Change

- [x] Documentation update

## Related Issue

Follow-on from #176. No new issue filed — the audit and the fix landed together.

### The seam

`neo.text_budget` holds three shapes, because the right cut depends on where the information is, and **picking wrong compiles, passes tests, and throws away the answer**:

| helper | keeps | for |
|---|---|---|
| `truncate_marked` | the head | source files, problem statements, formatted memory |
| `elide_middle` | both ends | tracebacks |
| `shown_of` | — | annotates an elided **list** |

`elide_middle` moved here from `engine`, where `persistent_reasoning` could not reach it without an import cycle.

### The two findings that were worse than the bug

**1. A nested cut emitted a marker that understated the loss by 29,110 characters.**

`_deliberate` cut each context section to 2,000 and the concatenation to 6,000. Measured:

```
verifiable_constraints: delivered 1,890 of 31,000 chars — its own marker sliced off
outer marker:           "... [truncated: 163 of 6163 characters not shown]"
```

A bare slice gives the model no information; this gave it **false** information in the confident register of an instrument reading. And it was never random which section lost — the order and caps were both fixed, so it was always `verifiable_constraints`, in the function used for the *hard* problems. A marker computed over a concatenated buffer is structurally incapable of describing source loss.

Sections now share the budget by max-min fair apportionment and are cut exactly once. That also recovered what the flat cap wasted — on a live store, a 30-char envelope plus 33,144 chars of memory against a 6,000 budget was sending **2,030 of it**; it now sends 6,000.

**2. `persistent_reasoning` used the head-cut helper on a traceback, and its own comment argued for the other one.**

```
tail-cut [:500] of a 40-frame traceback → 'ValueError: database is locked' present? False
```

That is the only line the root-cause analysis exists to find — and the cause it names instead is stored as a durable pitfall and replayed on later prompts.

Moving `elide_middle` exposed a latent bug in it: a fixed 1600-char head degenerates into a head-only cut at any budget below 1600, because `min(1600, 500)` claims the whole budget and starves the tail. The head is now a share of the budget; the existing traceback caller's output is byte-identical at its production budget, verified across input lengths.

### The rest

- **`engine`** — both exemplar sites appended a literal `...` whether or not anything was cut, so a 40-char solution was presented as continuing and a 4,000-char one was presented as continuing by exactly as much.
- **`pattern_extraction`** — cuts sit *inside* ``` fences, where a fence closing after a mid-function slice reads as a complete function. The extracted prevention rule is durable.
- **`constraint_verification`** — the caller treats the result as *the* constraint set, so a constraint in a dropped tail is not merely unextracted, it is unchecked.
- **`repair_loop`** — the input is malformed JSON, which is very often malformed at the *end*; a silent tail cut can remove the defect being repaired.
- **`algorithm_design`** — plus three silently elided **lists**, under a prompt that says "follow this exactly", so three bullets read as three edge cases.

The two prompts whose output is reused (`repair_loop`, `pattern_extraction`) now tell the model the marker is our note, not content.

## How Has This Been Tested?

- [x] `1990 passed, 8 skipped`; ruff clean
- [x] **Mutation-tested, not read.** Every claimed guard verified by breaking the code and confirming a test fails:

| mutation | caught |
|---|---|
| bare slice at each fixed site (incl. `[0:N]`, which slips past a grep) | ✓ |
| total cap raised 100× | ✓ |
| apportionment replaced by a flat per-section cap | ✓ |
| each of the three list-annotation slots removed | ✓ |
| traceback swapped to the head-cut helper | ✓ |
| exemplar reverted to the lying-ellipsis form | ✓ |
| separator reservation dropped | ✓ |
| `apportion` returning zeros instead of raising | ✓ |
| bare slice injected into an `engine` prompt | ✓ |

- [x] `apportion` fuzzed over **2,396,744** cases (0–6 sections × 8 sizes × 8 budgets): 0 violations — never over budget, never a negative or oversized allocation, never a phantom key, and never raising on a budget that *is* seatable
- [x] `elide_middle` diffed against the pre-move implementation at the production budget across 12 input lengths: 0 differing outputs
- [x] Reviewed by the `neo` CLI and an adversarial reviewer; the latter returned NEEDS WORK on round 1, and both headline findings reproduced before being fixed

**Test Configuration**: Python 3.13.9 · macOS 26.5.2 · neo 0.43.0

## Checklist

## Additional Notes

**Known limitation, documented rather than fixed:** `outcomes` truncates `diff_summary` at 2,000 with its own `... (truncated)` before `store` hands it to `pattern_extraction`, whose 500-char cut drops that inner marker — so the surviving marker describes the pre-cut value, not the original diff. Fixing it means changing what `outcomes` stores, which is outside this change. `text_budget`'s docstring warns to check whether a value arrives already cut.

**Not adopted:** the reviewing model argued the budgets should be measured in tokens rather than characters. True in principle, but it means a tokenizer dependency in a leaf helper and rewriting every caller's constants, and `ContextAssembler` already does token budgeting where the real budget lives. The helper is named honestly for what it measures.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -559,6 +559,22 @@
      files; `MAX_CHUNKS_PER_REPO` is fixed at 1000 while `--max-files` is not,
      so `files_with_chunks` reports the shortfall — `truncated` is False there,
      because the FILE cap genuinely was not the constraint.
+     **And the corollary the first version of this report broke: never blame a
+     cap for an absence it did not cause.** A selected file is missing from the
+     index for one of TWO unrelated reasons, and the console must not guess
+     between them. Either it produced no chunks at all — no function, class,
+     interface or struct for the grammar to match, as in an empty
+     `__init__.py`, an enum-only `.cs`, a type-alias-only `.ts` — which no cap
+     setting changes and which therefore gets a bare statement with NO remedy
+     attached; or the cap took every chunk it produced, which gets the cap
+     named and `lower --max-files`. `files_producing_chunks` is measured
+     BEFORE `_cap_chunks` precisely so the two stay separable; subtracting the
+     post-cap `files_with_chunks` from `selected` conflates them and is what
+     printed "the 1000-chunk cap is below the 25 files selected" for a build
+     that kept 559 of 559 chunks, with `chunks_capped` False on the same
+     report. It fired on this repo, on any repo with an `__init__.py`. A report
+     that invents a cause is worse than the silence it replaced, because
+     silence at least does not send the operator to the wrong knob.
   **Query footgun** (`language_parser.py`; incident detail in the tree-sitter
   doc's "Why queries break silently"): a query that fails to compile is
   indistinguishable from one that matches nothing — `_get_query` returns None
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -520,6 +520,58 @@
   and retrieval changes than waiting for an inference round trip. **Use it before
   believing any claim about what Neo "saw"** — the two defects below were both
   invisible from the outside and presented as the model being unhelpful.
+- Project index (`index/project_index.py`, `index/language_parser.py`; full
+  notes in `docs/tree-sitter-setup.md`). Three invariants, each of which was
+  violated and each of which produced an index that could not answer a question
+  about its own repository:
+  1. **Budgets are apportioned, never sliced.** `_select_files` groups eligible
+     files by language and hands each a share of `--max-files` proportional to
+     repo composition, floor of one slot per language (`_allocate_slots`); then
+     `_cap_chunks` round-robins `MAX_CHUNKS_PER_REPO` across FILES so each keeps
+     a chunk before any keeps a second. Both exist because a list slice is not a
+     ranking: globbing `**/*.py` before `**/*.cs` and slicing gave a .NET repo of
+     4,272 C# files an index of 83 Python files (95 from an in-repo worktree) and
+     zero C#, exit 0. Fixing only the file cut left `chunks[:1000]` re-creating it
+     one function later — chunks arrive grouped by file and files by language, so
+     the slice kept 1000 C# chunks and dropped every other language, with 37 of
+     the 100 selected files contributing nothing.
+  2. **Exclusion is two layers and `bin`/`build`/`out`/`target`/`dist`/`vendor`
+     belong to neither by name.** `EXCLUDED_DIR_NAMES` covers what repos forget to
+     ignore (`.worktrees`, `.claude`, `node_modules`, `obj`, virtualenvs);
+     `_build_exclusion_filter` layers the repo's own `.gitignore` on top via
+     `context_gatherer.load_gitignore_patterns`. **Footgun**: `should_ignore` only
+     tests the path handed to it, so ancestor directories must be walked
+     separately — `iter_paths` gets this via `os.walk` pruning, which the index
+     has no equivalent of. Matching is exact and case-sensitive, against
+     directory components only. The ambiguous names stay out because each is real
+     source somewhere (`src/bin/main.rs`, vendored trees; 254 tracked files under
+     `bin/`+`vendor/` across three local repos) and the asymmetry is one-sided:
+     over-excluding hides code permanently, over-including only spends slots.
+  3. **A cap that fired must be reported.** `selection_report` carries eligible /
+     selected / excluded / duplicates / chunk counts and the CLI prints them.
+     `truncated` means THE CAP BOUND US and is `examined < eligible` — whether
+     any candidate went unlooked-at, which the per-language iterators already
+     know. Both cheaper predicates name the wrong knob: `selected < eligible`
+     made dedup print "2 of 7 eligible files (capped at --max-files=1000)", and
+     `selected >= max_files` did the same whenever `--max-files` landed exactly
+     on the unique-file count. Raising a cap that never bound fixes nothing.
+     Separately, round-robin only represents every file while chunk slots ≥
+     files; `MAX_CHUNKS_PER_REPO` is fixed at 1000 while `--max-files` is not,
+     so `files_with_chunks` reports the shortfall — `truncated` is False there,
+     because the FILE cap genuinely was not the constraint.
+  **Query footgun** (`language_parser.py`; incident detail in the tree-sitter
+  doc's "Why queries break silently"): a query that fails to compile is
+  indistinguishable from one that matches nothing — `_get_query` returns None
+  and `parse_file` moves on; edge failures log at DEBUG. Four were broken at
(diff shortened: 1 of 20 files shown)


JLiotta-25:
**Note on CI:** `.github/workflows/ci.yml` triggers only on PRs targeting `main`, so a stacked PR gets no checks. CI will run automatically once #177 merges and GitHub retargets this to `main`.

Covered locally in the meantime:
- Full suite on 3.13: `1990 passed, 8 skipped`; ruff clean (via the repo's pre-commit parity gate on both commits)
- All 9 changed files compile under 3.11 and parse against the **3.10** grammar (the declared floor)
- No 3.11+/3.12+-only syntax or stdlib in the diff

That is not a substitute for the 3.10–3.13 matrix, which the repo's own hook is explicit about. Worth a second look after retarget.

mliotta (commented):
Reviewed by checking the branch out and running it. The seam is right, the nested-cut finding is real and correctly diagnosed, and the mutation testing is the strongest evidence any PR here has carried. `84 passed` on the two new files locally.

One gap in the completeness claim, and it is the same structural defect the PR itself identifies elsewhere. Not a blocker for the mechanism — a blocker for the word *every*.

## `_community_fallback_learnings` still cuts twice, silently, straight into the planning prompt

`engine.py:2346-2351` — a file this PR already edits, in the section this PR already apportions:

```python
picked = [f for h, f in scored if h > 0][:2] or [f for _, f in scored[:2]]
...
lines = ["## Community Patterns (fallback)"]
for f in picked:
    body = (f.get("body", "") or "")[:200]
    lines.append(f"- **{subject}**: {body}")
```

Both defect shapes the module exists to remove: a bare `[:200]` on text, and a silent `[:2]` on a list. Run against a cache of five 912-character facts:

```
## Community Patterns (fallback)
- **Pattern 0**: RULE: xxxxxxxxxx…      <- stops dead at 200 chars

has truncation marker: False
has list annotation  : False
facts in cache: 5 | facts shown: 2
body chars per fact: 912 | shown: 200
```

It reaches a model by two paths, both of which this PR touches:

- `engine.py:1712` sets `past_learnings = [fallback]`, and `engine.py:1764` passes `past_learnings` into `_format_planning_prompt`.
- `past_learnings` is one of the three `_DELIBERATION_KEYS`, so it is also a section `_deliberation_context` now apportions.

Which makes the second path the PR's own headline finding, one function over. `_community_fallback_learnings` cuts each body to 200 with no marker; that becomes `past_learnings`; `_deliberation_context` then cuts *that* with a marker describing the already-cut value. A marker over a pre-cut input reports the wrong denominator — exactly the `outcomes` → `pattern_extraction` nesting the PR documents as a known limitation, except here both halves are in `engine.py` and the outer half is new code in this diff.

The fix is two calls to helpers this PR already ships: `truncate_marked(body, _COMMUNITY_BODY_CHARS)` and `shown_of(candidates, len(picked))` on the header. The list one matters on its own terms — these are curated global rules, and "here are the community patterns" over a silently-halved list is the "three bullets read as the complete set" case the module docstring names.

Worth a pass over the audit that produced *seventeen*: this one is a `[:200]` on a `.get()` result, so a grep for slices on named variables would miss it, and it sits in a fallback branch that only fires when `FactStore` retrieval misses — which is also when the model has the least other context and leans hardest on what it is given.

## Minor

**`elide_middle` can exceed its budget in the degenerate case.** `elide_middle('x'*5000, 10)` returns 32 characters — the bare marker, over a budget of 10. Defensible (an honest marker beats silence), and unreachable from the one caller at 4000, but the docstring promises `budget` and this is the one path that does not honour it. Worth one line saying so, given the module's whole thesis is that quiet contract violations are what bite.

**`apportion` leaves up to `len(sections) - 1` characters unspent** in the all-hungry branch — `apportion({a,b,c: 10000}, 100)` allocates 99. Rounding, harmless, and the comment already says "split it evenly and stop". Noting only because the fair-share rationale is otherwise about not wasting budget.

**`plan_summary` (`engine.py:1409-1411`)** does `[:80]` per step and `[:300]` on the join, unmarked, into `persist_simulation_episode`. Not prompt-bound today, so out of scope — but persisted episodes are retrieved and surfaced into later prompts, so it is the same class one hop removed. Mentioning it so the next audit does not have to rediscover the hop rather than suggesting it be fixed here.

## What is right

`apportion` is the correct generalization and the 2,030-of-6,000 measurement justifies it far better than a per-section cap ever could. Splitting the three shapes by *where the information is* — and saying plainly that picking wrong "compiles, passes tests, and throws away the answer" — is the part that will keep this from regressing, more than the tests will.

The mutation table is the right way to prove a guard exists, and including `[0:N]` because it slips past a grep is the detail that shows the audit was done rather than claimed. Verified independently: `elide_middle` moved with no behaviour change, and the empty-text-at-negative-budget branch in `truncate_marked` is a real bug caught, not a hypothetical — `budget >= len(content)` is `-1 >= 0`, false, so it would have emitted "0 of 0 characters not shown" from the one function whose contract is that a marker means a real cut.

Declining the tokens-over-characters suggestion is the right call for the stated reason, and naming the helper for what it actually measures is the honest version of that decision.

---

Note the base moved under this PR: #177 now carries two more commits (a report line that blamed the chunk cap for absences it did not cause, and C# `qualified_name` bases). Merges clean — verified with `git merge-tree`, exit 0.

mliotta (approved):
Approving. The gap my review found — `_community_fallback_learnings`' unmarked body cut and silent list elision, both feeding `past_learnings` and therefore nested under this PR's own new apportionment — is fixed in `bed71d6` on this branch, with six tests (four fail against the pre-fix source, plus the AST helper check once `shown_of` is required of `neo.engine`).

The three minor notes are left as-is and are not worth holding this: `elide_middle`'s degenerate over-budget return, `apportion`'s sub-`len(sections)` rounding remainder, and `plan_summary`'s unmarked cut into `persist_simulation_episode` (not prompt-bound today, so a later concern).

The seam is the right one, and the mutation table is the strongest evidence any PR in this repo has carried.

mliotta (approved):
Approving — the review gap is fixed in `bed71d6` on this branch with six tests, and CI is green on all four Python versions.

PR #186, opened 2026-08-10
fix(gatherer): implement .gitignore instead of approximating it

JLiotta-25:
## Description

`context_gatherer.iter_paths` runs on **every** neo invocation, not just `--index`. On a repo containing nested checkouts and agent worktrees, a single prompt selected 16 files of which **12 were worktree copies** — the same file six times, at identical scores — and 14 of 16 were paths the repo's own `.gitignore` excludes.

#159 fixed this class in `ProjectIndex`. I assumed it was an indexing problem. It wasn't: the gatherer had the same defect and none of the same defences.

## Type of Change

## Related Issue

Follow-on from #159. Found by measuring a live repo rather than from a report.

### `should_ignore` did not implement `.gitignore`; it approximated it

Six defects, each of which silently changed what the model was allowed to read.

**1. Unanchored directory rules did not match at depth.** `if pattern.endswith('/')` won the `if/elif` chain and both clauses inside it were root-anchored, so `build/` never reached the match-at-any-depth logic and was treated as if it were `/build/`.

```
build/         vs src/build                → False   (git: True)
node_modules/  vs pkg/node_modules/x.js    → False   (git: True)
obj/           vs src/Api/obj/Debug/x.dll  → False   (git: True)
```

That is the form **457 of 565** directory rules in this workspace take. The anchored form — 108 lines, and **100 of them in a single repo** — is what an earlier cut of this PR fixed. I had generalised from the least representative repository on the machine.

**2. Negation was skipped entirely.** Returning on first match makes `!` unreachable by construction. One repo's `.gitignore` pairs `.claude/*` with `!.claude/skills/`; skipping the negation hid **7 git-tracked files** that `git check-ignore` reports as *not* ignored. Now last-match-wins, per gitignore.

**3. `*` crossed path separators.** `fnmatch` is component-blind, so `/*.png` — "a PNG at the repository root" — matched `docs/audits/…/img/concept-5-b.png` six directories down. The new `_path_glob` compiles to a regex where `*` is `[^/]*`, `**` spans components only when it is a whole component (git treats `a**b` as a plain star), and a leading `**/` is `(?:.*/)?` so it matches at the root too.

**4. Character classes were mistranslated, in two ways that cancel out of a casual reading.** `[!a-z]` was compiled as a *positive* class — the exact inverse of its meaning — and only on the anchored side, so identical syntax meant opposite things depending on whether the pattern happened to contain a slash. A class could also consume `/`, letting one component's rule reach across a separator, which is defect 3 by another route.

**5. A malformed pattern crashed the process.** `[]]` is the POSIX spelling for "a class containing a literal `]`". Scanning to the first `]` cut it into an empty class and left the regex unbalanced, raising an uncaught `re.PatternError`. Since `should_ignore` runs on every directory and every file of every walk, one such line in any `.gitignore` turned every neo invocation into a traceback — where the `fnmatch` it replaced could only ever degrade to a non-match.

**6. A bare directory name was tested against the basename.** `node_modules` never matched `a/node_modules/b.js` — the walk hid it by pruning the directory first, which means the predicate was only correct while its caller compensated. Splitting on `/` alone also broke pruning on Windows, since `iter_paths` builds candidates with `os.path.join`.

### Exclusions: one list, and named by layout

`EXCLUDED_DIR_NAMES` is **deleted**. 27 of its 31 names were already in the shared defaults, and duplication cost exactly what duplication costs — when the gatherer's copy was corrected, the index's copy kept hiding the same files and the two subsystems disagreed about `.claude` until someone diffed them.

The worktree rule is written as a **layout** (`**/.claude/worktrees`), not as either bare component, because both shortcuts hide committed source:

| shortcut | what it hides |
|---|---|
| `.claude` / `.codex` / `.car` | 60 tracked skill implementations across two repos, e.g. `.claude/skills/deploy-app/scripts/deploy_verify.py` |
| bare `worktrees` | `src/worktrees/manager.ts` — a worktree **manager** keeps source in a directory named for what it manages |

Both are the same error, and the same one as excluding `bin/` in #159: excluding a container for what sometimes sits inside it.

## How Has This Been Tested?

**Validated against `git check-ignore` itself** — the only oracle that settles gitignore semantics — in **both** directions:

```
7,534 on-disk paths across 33 real repositories
  → 0 over-exclusion
  → 2 under-exclusion, both from nested .gitignore files
```

Nested `.gitignore` files are not read and never were; that limit is pre-existing and unchanged by this PR.

An earlier revision of this description claimed `1,709 tracked files → 0 disagreements`. That oracle was much weaker than it looked and has been replaced: **git does not apply `.gitignore` to tracked files**, so without `--no-index` every expected answer was "not ignored" and the under-exclusion direction — the one this PR exists to fix — was unmeasurable by construction.

- [x] `2116 passed, 8 skipped`; ruff clean
- [x] The original defect, on the live repo: worktree files **12 → 0**, gitignored **14 → 0**, duplicate slots **9 → 0**
- [x] No over-exclusion of the contested directories: `forge`, `car`, `parslee-knowledge` — 42 tracked files, **0 hidden**
- [x] 15 malformed / edge-case patterns (`[]]`, `[a-]`, `[z-a]`, `[[]`, `*[`, `***`, `!!x`, `\`, …) each asserted against the value `git check-ignore` actually returns, plus a walk-level test that one bad line no longer aborts gathering
- [x] Mutation-tested: reverting each fix fails its guard

## Checklist

- [x] My changes generate no new warnings — including a `FutureWarning: Possible nested set` that `[[]` used to print on every invocation

## Additional Notes

**Scope grew, deliberately.** This started as "honor anchored patterns" and became "implement gitignore" because each fix exposed the next, and every one was measured on real repositories rather than argued. A change whose thesis is *honor `.gitignore`* cannot ship ignoring the rule that rescues files.

**Reviewers disagreed on structure, and it is worth flagging.** Neo recommended splitting this into two PRs (the incident fix vs. the gitignore-semantics work) and dropping the ~20 added default names as scope creep. I kept them in one change: those names are `EXCLUDED_DIR_NAMES`' existing policy, which the index already applied, and dropping them would mean either regressing the index or keeping the two lists that caused the divergence in the first place. Linus reviewed the structure and found it right. Both reviewers' must-fix findings are addressed; neither has re-reviewed this final state.

**One part was dropped after measuring.** Content dedup in the gatherer was planned; the exclusion fixes took duplicate slots to 0 on their own, so it would have been machinery for a problem already gone.

**Known limits, unchanged:** nested `.gitignore` files are not read (2 of 7,534 paths). The defaults are additive over `.gitignore` with no escape hatch — `--include` filters survivors and cannot un-prune a directory — which makes each default a one-way decision, and is why the two contested entries were narrowed rather than kept.


kkeeling (approved):
Approved by Keenan via Kai — merge authority granted 2026-08-11 (unified-store hill-climb, plan doc landing next). Forensic assessment: docs/audits pending.

Issue #183, opened 2026-08-09
Learning benchmark gates on wall-clock latency, so CI fails randomly on a slow runner


mliotta:
`test_benchmark_passes_all_required_scenarios_and_safety_gates` asserts `report.accepted is True`, and acceptance includes a **wall-clock** gate:

`benchmarks/learning_loop_v1.json`
```json
"safety_thresholds": {
  "harmful_memory_rate_max": 0.0,
  "unsupported_promotion_rate_max": 0.0,
  "repeat_error_rate_max": 0.0,
  "project_leakage_rate_max": 0.0,
  "latency_ms_max": 500.0
}
```

Observed on CI (PR #180, `test (3.10)`), with **all twelve scenarios `passed=True`** and every correctness rate exactly at its threshold:

```
accepted=False
acceptance_failures=['latency_ms=592.4409 exceeds latency_ms_max=500.0000']
```

Nothing about the run was wrong. The runner was slow.

## Why this will keep happening

Same benchmark, same commit, on a developer machine:

```
run 1: latency_ms=53.8  accepted=True
run 2: latency_ms=53.1  accepted=True
run 3: latency_ms=53.4  accepted=True
```

53ms locally against a 500ms gate looks like 9x headroom. The CI runner measured 592ms — **11x slower than local**, so the headroom is not real. GitHub's shared runners vary by more than the margin the threshold assumes, and nothing in the benchmark controls for that.

The blast radius is every PR: a failure here reads as `1 failed, 1761 passed` with a correctness-shaped assertion (`assert report.accepted is True`) and a 9,000-character repr, so the first read is "my change broke the learning benchmark". It cost a real detour to rule that out.

## The shape of the problem

`latency_ms` is the only machine-dependent value in `safety_thresholds`. The other four are deterministic rates that are exactly `0.0` on any machine — they are properties of the algorithm. Latency is a property of the hardware, and grouping it with them means a slow runner is reported in the same register as a memory-safety violation.

`accepted` is also the benchmark's *published* verdict, not just a test assertion, so a timing wobble currently invalidates a correctness claim.

## Options

1. **Separate the gates.** Keep `accepted` for correctness and safety; report latency as a metric with its own advisory verdict. The test then asserts what it means to assert, and a slow runner does not invalidate a correctness claim. Preferred — it is the only option that fixes the category error rather than the symptom.
[...]

PR #187, opened 2026-08-10
fix(dry-run): show what would actually be sent, and stop the panel billing you

JLiotta-25:
## Description

`--dry-run` exited in `cli.main` **before the engine was constructed**. It printed the gathered file list and nothing else — while `CLAUDE.md` told operators it "assembles the full context (file selection, fact retrieval, constraints, four-layer assembly) and prints what *would* be sent to the LM." Three of those four never ran.

That matters more than an ordinary documentation slip, because this is the instrument the project tells people to use *before believing any claim about what Neo saw*. It is how the gatherer defect in #186 was found. Invisible through it: the Execution Envelope (which reaches the model through seven prompt builders), every retrieved fact, and the REPOSITORY CONTEXT block carrying the #178 truncation markers — the markers whose entire purpose is that a cut be visible could not be seen through the tool built for seeing.

## Type of Change

- [x] Documentation update

## Related Issue

Follow-on from #158, #178 and #186. Found by auditing where else Neo decides what the model sees with no signal when it's wrong.

### `--dry-run` could make real, billed LLM calls

The worst defect, and the one that survived two reviewers before being caught. The multi-agent panel does **not** reason through `self.lm`: `_build_car_role_factory` calls `create_adapter("car", model=m)` per role and uses `self.lm` only as the fallback for unassigned roles. So installing a recording adapter did not intercept it.

With `car-server` reachable — this project's documented normal setup, since the observer autostarts off it — a novel prompt under `--dry-run` ran the full panel against real models, spent real money, never raised the stop signal, and printed ordinary output. The flag was a no-op that bills you. **Measured: 4 real adapters built (planner, coder, critic, judge); now 0.**

It is also the honest scope. The panel's later prompts are built from earlier model responses, so they cannot be shown without making the calls the flag exists to avoid.

### The prompt is recorded, never rebuilt

`RecordingLM` is a real `LMAdapter` in the engine's own `self.lm` slot, so the output is the exact `messages` list Neo hands the adapter. A renderer walking the context dict would be a second implementation of the seven prompt builders, free to drift the moment one changed — the duplicated-rule shape that put `EXCLUDED_DIR_NAMES` in two places and cost a release.

`DryRunComplete` derives from `BaseException` so `_process_guarded`'s `except Exception` cannot report a dry run as a `FAILED` lifecycle event. An ordinary exception is **not** a safe substitute: `_deliberate` has its own `except Exception` that would swallow it and silently degrade to the fast path.

### Three claims narrowed to what actually holds

| claim | why it was false |
|---|---|
| "the exact prompt that would be sent" | Anthropic hoists `system` into a separate kwarg, Google remaps roles and content keys, Ollama flattens to one string, CAR adds `intent_json`. And no provider is resolved at all, because the flag deliberately requires no credentials — so it structurally cannot render the wire payload. |
| "mutates nothing" | `FactStore.initialize` runs prune → demote → purge and then **saves**; `demote_unhelpful_facts` lowers confidence and invalidates facts, all pre-inference. `detect_implicit_feedback` was not "the one" such call, only the one I had found. |
| model params as if sent | Anthropic drops `temperature` on Opus 4.7+/Sonnet 5, CAR lets backend defaults win, `reasoning_effort` reaches only the OpenAI-family adapters. |

`FactStore(read_only=True)` makes `save()` a no-op at the single write choke point — a new caller cannot forget it, which threading a flag through every retrieval path could. `metrics.jsonl` still records, deliberately: the two events a dry run writes are read by neither `citation-stats` nor `learning-stats`.

### Three more defects found while moving the stop into the engine

- **`--stdin-json --dry-run` never stopped at all.** The old early-exit lived inside the plain-text branch, so JSON callers made a real inference call under a flag whose whole promise is that it does not.
- **`--json --dry-run` broke both of its documented invariants**: zero documents on stdout where the contract promises exactly one, no terminal event so `reasoning` stayed `running` forever, and every source line beginning with `{` became a counterfeit event for a host parsing the JSONL stream by that prefix.
- **`--no-scan` left `gathered` unbound**, and `--dry-run` briefly began requiring an API key once the stop moved past adapter resolution.

### The Execution Envelope's cuts are now marked

`prompt_section()` had three bare slices — `constraints[:12]`, `success_criteria[:8]`, `attempts[-3:]`. Under a prompt saying "satisfy these constraints", twelve of thirteen read as thirteen. The #178 sweep missed this module for the reason that sweep wrote down about itself: it looked for slices inside the known prompt builders, and this is a method on a dataclass reached indirectly through `_retrieve_context`.

Caps are unchanged; only the loss is visible. `shown_of(..., tail=True)` renders `[showing last 3 of 20]` for the tail slice, because which end survived is part of the meaning — a loop that has run twenty times is not a sample of twenty. `retrieval_query()` keeps its bare slices deliberately: it is embedded and compared by cosine, never read, so a marker there would be tokens in the query vector.

### Two invariants that were held by comments only

Both are now tested, because a comment is exactly what held `EXCLUDED_DIR_NAMES` in two places while one copy was corrected and the other kept hiding files.

- **The ranking formula stays single-sourced.** Checked: both retrieval paths already call `rank_score` and nothing recomputes it inline. A differential over 60 facts plus an AST walk for stray `success_bonus`/`provenance_bonus` calls now keeps it that way.
- **Every declared tree-sitter query matches real code.** java, go, c, ruby and php had no corpus in the checkout, so they had only ever been compile-tested — the exact gap #158 lived in.

Two coverage gaps are **pinned rather than closed**, because a documented gap beats a silent one: ruby declares no edge queries at all, and php has imports but no inheritance, so `class X implements Y` yields an edge in java, c_sharp and typescript and nothing in php.

## How Has This Been Tested?

- [x] `2119 passed, 8 skipped`; ruff clean
- [x] Panel suppression verified by patching `create_adapter` and forcing `_car_route_capability` to report a full pool: **4 adapters built before, 0 after**
- [x] `facts_global.json` byte-identical across two dry runs; `--json` stdout parses as exactly one document with 1 terminal event and 0 counterfeit events
- [x] Mutation-tested: the three `shown_of` markers, the `detect_implicit_feedback` guard in both directions, and **17 of 17** single-query mutations in the language-yield tests (the per-language form these replaced caught **0 of 28**)
- [x] Two tests that could not fail were found and fixed — a `getattr` chain on attributes `CodeEdge` does not declare, and per-language assertions that let a working query mask a dead one
- [x] Round 3: the VERIFY dry-run path wrote **1 episode + 1 `citation_survival`** and emitted **0 stdout documents** under `--json`; now 0, 0 and one valid document. Both `_get_query` mutations that break real extraction now fail (they left all 29 green)

## Checklist

## Additional Notes

**Independent of #186** — no file overlap, so the two can merge in either order.

**Reviewers disagreed once, and Linus was right.** Neo proposed making `DryRunComplete` an ordinary `Exception` caught explicitly before the broad handler. That would have been swallowed by `_deliberate`'s own `except Exception` and silently degraded every dry run to the fast path. `BaseException` stays.

**Three review rounds, and every round found defects in the seams the previous round created** — not in what it fixed. Round 3's were: a dry run writing to the episode ledger via VERIFY mode's normal return, `--json` fixed on one exit and not its twin five lines away, and a test file that routed around the very accessor it was testing.

**Some of it was invisible to me locally.** The panel billing bug was undetectable here because `car-server` isn't reachable on this machine, so that path was dead in every test I ran. And a claim I wrote in round 2 ("the two events it writes are read by neither citation-stats nor learning-stats") was measured on the fast path and generalised to a path that did not exist when I measured it.

**Known limits, stated rather than fixed:** the output is the adapter's input, not the wire payload; only the first call is shown, and the panel is suppressed rather than previewed; `metrics.jsonl` still grows.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -551,12 +551,33 @@
   `pi_boost + hist_boost + _symbol_score` and applies an adaptive limit and a byte
   budget. A first-pass harness overstated R@10 by 0.14 against the real CLI. Validate
   any in-process replica against `--dry-run` output before trusting a sweep.
-- Debugging: `neo --dry-run "your query"` assembles the full context (file selection,
-  fact retrieval, constraints, four-layer assembly) and prints what *would* be sent to
-  the LM, then exits without making the LLM call. Faster iteration on context-gatherer
-  and retrieval changes than waiting for an inference round trip. **Use it before
-  believing any claim about what Neo "saw"** — the two defects below were both
-  invisible from the outside and presented as the model being unhelpful.
+- Debugging: `neo --dry-run "your query"` runs the real engine — file selection, fact
+  retrieval, constraints, four-layer assembly — and prints the **exact messages** that
+  would go to the provider, then exits without making the LLM call. Faster iteration on
+  context-gatherer and retrieval changes than waiting for an inference round trip.
+  **Use it before believing any claim about what Neo "saw"** — the two defects below
+  were both invisible from the outside and presented as the model being unhelpful.
+  This bullet described the tool's *intent* for a long time and not its behaviour: the
+  flag used to exit in `cli.main` **before the engine was constructed**, so three of
+  the four things listed above never ran and the output was the file list alone. The
+  Execution Envelope, retrieved facts, and the REPOSITORY CONTEXT block with its
+  truncation markers — the #178 work, whose entire point is that a cut be visible —
+  were all uninspectable through the tool built for inspecting them. An instrument
+  that under-reports sends the operator to the wrong knob, which is the same failure
+  as a cap that blames itself for an absence it did not cause.
+  **The prompt is recorded, never rebuilt** (`neo.dry_run.RecordingLM` is a real
+  `LMAdapter` installed in the engine's own `self.lm` slot), because a renderer that
+  walked the context dict would be a second implementation of the seven prompt
+  builders, free to drift the moment one changed — the duplicated-rule shape that put
+  `EXCLUDED_DIR_NAMES` in two places. `DryRunComplete` derives from `BaseException`,
+  not `Exception`: `_process_guarded` converts anything its `except Exception` catches
+  into a `FAILED` lifecycle event, and reporting a dry run as a crash would be one
+  more way of misdescribing the run. **A dry run mutates nothing** — the old
+  implementation had that for free by never reaching the engine, so `dry_run=True`
+  now skips `detect_implicit_feedback`, the one pre-inference call that both changes
+  confidence and saves. Pinned by `test_dry_run_skips_the_one_call_that_saves`; the
+  file-diffing test beside it is a backstop only and stays green either way, because
+  that call is a no-op without a prior session.
 - Project index (`index/project_index.py`, `index/language_parser.py`; full
   notes in `docs/tree-sitter-setup.md`). Three invariants, each of which was
   violated and each of which produced an index that could not answer a question
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -597,15 +597,26 @@
   reaching the engine at all meant a "read-only" inspection was aging the store it
   inspected. `FactStore(read_only=True)` makes `save()` a no-op at the single write
   choke point, which a new caller cannot forget; `dry_run` also skips
-  `detect_implicit_feedback`. Retrieval still marks facts accessed in memory, and
-  `metrics.jsonl` still records the run — deliberately: the two events it writes
-  (`execution_context_resolved`, `retrieve`) are read by neither `citation-stats`
-  (which filters `citation_survival`) nor `learning-stats` (which reads the episode
-  ledger), so observability costs nothing here.
-  Under `--json` the report is the single stdout document and a terminal `completed`
-  event is emitted, because writing prose to stderr broke both `--json` invariants at
-  once: zero documents on stdout, and every source line beginning with `{` became a
-  counterfeit event.
+  `detect_implicit_feedback` and `_complete_learning_episode`. Retrieval still marks
+  facts accessed in memory, and `metrics.jsonl` still records the run —
+  deliberately: the two events it writes (`execution_context_resolved`, `retrieve`)
+  are read by neither `citation-stats` (which filters `citation_survival`) nor
+  `learning-stats` (which reads the episode ledger), so observability costs nothing.
+  **That argument was measured on the fast path and briefly untrue on another.**
+  VERIFY mode reasons without an LM call, so `process()` returns NORMALLY and never
+  raises `DryRunComplete`; `_complete_learning_episode` then wrote an episode file
+  AND a `citation_survival` metric — the exact two surfaces the sentence above
+  claims are untouched. Measured on a clean HOME: 1 episode, 1 `citation_survival`.
+  Gating that call is what makes the claim true; do not narrow the gate to the
+  recorded-call path.
+  Under `--json` the report is the single stdout document (`{dry_run, calls, note}`
+  — a second schema, discriminated by `dry_run: true`, with no `orchestrator` key;
+  `test_host_adapter_parity.py` does not know about it) and a terminal
+  `phase_completed(reasoning)` + `completed` pair is emitted. Writing prose to
+  stderr broke both `--json` invariants at once: zero documents on stdout, and every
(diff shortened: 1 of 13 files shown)


kkeeling:
## Rebased onto `main` (`68c197e`, post-#194 and post-#202), tests green

**Rebase.** Four commits replayed, Jennifer's authorship preserved on all four. Two conflicts, both semantic, both resolved by keeping *both* sides rather than picking one:

**1. `src/neo/execution_context.py` — `prompt_section()`.** #200 (proof-aware execution) added Validation gates / Validation evidence / Hypotheses blocks to the same function this branch rewrites, and added them as bare `[:12]`, `[:12]`, `[:5]` slices. Kept this branch's `shown_of`-marked `Success criteria` line *and* #200's three new blocks — and marked those three too, with named `_MAX_VALIDATION_GATES` / `_MAX_VALIDATION_OBSERVATIONS` / `_MAX_HYPOTHESES` constants at the same values. Caps unchanged; only the loss is now visible.

That last part is a judgement call, so here is the reasoning: the function's docstring, as rewritten by this branch, says *"Every cut here is MARKED."* Merging #200's three unmarked slices underneath it would have made the docstring false on the merge commit itself — the precise failure this branch exists to remove, arriving through the merge rather than through an edit.

**2. `CLAUDE.md`.** #194 added a "Measuring retrieval changes" bullet immediately above the `--dry-run` bullet this branch rewrites. Both bullets kept.

### Test reconciliation (`ea990de`, mine, not Jennifer's)

Two of this branch's own guards broke on contact with `main`. Neither is a behaviour change:

- **`tests/test_execution_envelope_truncation.py`** — `ResolvedExecutionContext` gained four required fields in #200, so the `_context()` helper could no longer construct one. Supplied. The generic `TestNoBareSlicesRemain` guard is also extended to the three lists #200 brought: that guard's stated premise is *"adding a new capped list without a marker is the recurrence this file exists to catch"*, so leaving them out would have let the guard pass while the defect it names sat three lines below the ones it checked.

- **`tests/test_rank_mine_eval.py::TestOutputFormatContract`** — #194's format-contract guard recovered the CLI's f-string by scanning source **lines** for one holding both `print(` and `bytes (score:`. This branch moved that print into `_print_selected_files` and wrapped the literal across two physical lines, leaving `bytes ` on one and `(score:` on the next. **The format string did not change a character**, and the guard reported *"the parser's contract has moved and this test is blind."* It now reads the template off the AST, where implicit f-string concatenation is a single `JoinedStr` node — a reflow is invisible, a real format change still fails loudly. Verified end-to-end: `neo --dry-run` still emits `  <path> (lines a-b) - <n> bytes (score: x.xx)`, so `tools/rank_mine_eval.py` and `tools/m2_battery.sh` both keep parsing.

### Tests

`pytest` on the rebased branch: **2313 passed, 8 skipped, 2 failed** in 2250 s. Both failures are pre-existing and reproduce on `main` with this branch's changes absent:

| test | branch | main (post-#194) | cause |
|---|---|---|---|
| `test_integration.py::…::test_consolidation_performance_with_all_boosts` | 173 ms vs 100 ms limit | **202 ms**, same machine | wall-clock assertion used as a correctness gate — the #183/#185 defect class, surviving in a different file |
| `test_operating_modes.py::…::test_verify_cli_needs_no_provider_and_agent_cli_fails_before_provider` | 30 s subprocess timeout | **same timeout**, same machine | the subprocess runs under `HOME=tmp_path`, so a cold fastembed cache is re-fetched inside the 30 s budget |

Neither reproduces on the runner — `main` is green — so CI is the gate here, not the local box. Both were re-run in isolation to rule out load from the parallel M1 measurement.

<sub>Goal 2 of `docs/unified-store-plan.md` — Land the queue. #194 merged as `281cf87`; #190 rebases onto this branch next.</sub>

kkeeling (approved):
Rebased onto `main` (`68c197e`, post-#194 and post-#202). Two semantic conflicts, both resolved by keeping both sides — full reasoning in the evidence comment. CI green on 3.10-3.13 plus the new guard-invariant battery (49s).

The `--dry-run` output format is unchanged end-to-end, so `tools/rank_mine_eval.py` and `tools/m2_battery.sh` keep parsing it; #194's format-contract guard now reads the template off the AST rather than off source lines, which is what a reflow should cost.

Approving under the merge authority recorded in `docs/unified-store-plan.md` (Keenan, 2026-08-11).

PR #188, opened 2026-08-10
fix(gatherer): demote test files on the keyword path, not just the semantic one

JLiotta-25:
## Description

Asked to fix test/doc dominance in file ranking, I measured three candidate fixes. **Two were rejected by the measurement and one shipped** — the rejections are the more useful half of this PR, because each was a change I had already argued was correct on principle.

Measured symptom: "add retry logic to the CAR adapter" ranked `src/neo/adapters.py` **14th**, below nine test and doc files, chunk-split.

## Type of Change

## Related Issue

Follow-on from the audit that produced #186 and #187.

### What shipped: `TEST_PENALTY` applies to the keyword path too

It already existed, and already fixed this — on the ProjectIndex semantic-boost path only, declared inline inside `gather_context`. **The two scoring paths disagreed about test files and only one was ever corrected**, which is the same duplicated-rule shape that let `EXCLUDED_DIR_NAMES` hide files from one subsystem while the other was fixed. There is now one constant, one `is_test_path`, used by both, and a test that greps for a second definition.

Why it matters: a test file's name is a strict **superset** of its subject's filename tokens — `test_car_adapter.py` matches `{adapter, car}`, `adapters.py` matches `{adapter}` — so on the keyword path it can only ever score at least as high as the code it tests, and the implementation then takes a size penalty for being large.

| corpus | metric | before | after |
|---|---|---|---|
| 12 code prompts | mean rank of first source file | 4.50 | **2.00** |
| 12 code prompts | mean tests+docs in top 10 | 5.58 | **4.25** |
| 6 doc-seeking prompts | mean docs in top 10 | 2.16 | **3.16** |

Both prompt classes improve. Treat the numbers as directional — self-chosen unlabelled prompts, one Python repo, and a mean over an unbounded rank is dominated by its worst case.

### What was rejected, and why that matters more

**Document-frequency filtering of prompt tokens.** `a` appears in 49 of 85 source basenames, handing each +0.60 and consuming one of only three `min(hits, 3)` slots. A real defect, and the same fix `select_chunks` already has one level down. Removing it made ranking **worse**: 4.50 → 5.50 first-source, 5.58 → 5.75 tests+docs. Only 2 of 12 prompts moved and both degraded — which supports "no measurable benefit at n=2", not the stronger causal claim I first wrote. Dropped either way.

**Cutting the blanket `docs/` +0.8 bonus to +0.2.** Helped code prompts (4.50 → 2.83), cost doc-seeking prompts (2.16 → 0.66). A direct trade between two prompt classes, declined.

### Two defects found in the process

**The escape hatch was a substring match.** `prompt_is_test` included `"spec"`, so **"in-SPEC-tor"** classified "the A2UI inspector shows a stale fact count" as a testing prompt and every test file kept full score. Survivable while it only softened a cosine; not once it gated the keyword path — that one prompt moved the 12-prompt mean by itself. Now word-boundary anchored, with `\btest` as a prefix so "testing" matches and "latest" does not.

**`is_test_path` was Python-only, and I shipped it in the same commit that fixed the identical bug twelve lines below.** It was `rel_path.startswith("test")` — unanchored, no separator. Wrong on 9 of 11 real conventions:

```
MISS  pkg/foo_test.go     Go        OVER  testdata/schema.py   real source
MISS  src/Foo.Tests/*.cs  .NET      OVER  testing/harness.py   real source
MISS  src/foo.test.ts     Jest      OVER  testbed/server.py    real source
MISS  spec/*_spec.rb      RSpec
MISS  FooTest.java        JUnit
```

It looked correct only because the corpus is one Python repo, where it matches 99 of 295 files with no false positives. Now 16 of 16.

### The penalty re-ranks; it must never evict

This was never only a ranking change. Scaling bonuses can push a file under `MIN_SCORE_THRESHOLD` and drop it from context **entirely** — a one-hit test file at depth 1 went 0.55 → 0.19 against a 0.2 floor. And both metrics above improve monotonically when a file *disappears*, so neither instrument could see the cost it was paying.

`score + forfeited` reconstructs the undemoted final score exactly (every step after the multiplier is additive), so a file that would have been admitted stays admitted, ranked beneath everything else. The floor lands at the **end** — applied at the multiplier it is eaten by the depth penalty, 0.24 → 0.19, straight back under.

## How Has This Been Tested?

- [x] `2082 passed, 8 skipped`; ruff clean
- [x] Four mutants caught: penalty removed, escape hatch stuck true, stuck false, `is_test_path` stuck false
- [x] `is_test_path` asserted against Go, .NET, Jest, Angular, RSpec, JUnit and Python conventions, plus the three false positives
- [x] Eviction pinned in both directions: an admitted file stays admitted; a file that would have been dropped is not rescued

## Checklist

## Additional Notes

**One reviewer must-fix was declined, with evidence.** Giving the identifier regex the same `{3,}` floor as its sibling would drop `db`, `os`, `ui`, `id`, `fs` — the short identifiers CLAUDE.md documents as carrying the most signal, and the reason that file chose document frequency over length. Measured: a length floor has no demonstrated ranking benefit and a documented cost.

**Known limit, stated rather than fixed:** `"add a test for the retry logic in the CAR adapter"` correctly trips the escape hatch and turns the demotion off — restoring the old ranking on a prompt that needs *both* files. Not a regression (main behaves the same), but the win is not unconditional.

### DIFF
--- a/src/neo/context_gatherer.py
+++ b/src/neo/context_gatherer.py
@@ -413,17 +413,30 @@ def score_candidate(rel_path: str, size: int, prompt_tokens: set[str],
     only ever outrank the code it tests, and then the implementation takes a
     size penalty on top.
 
-    Measured over 12 self-chosen code prompts on THIS repo: the first source
-    file ranked 4.50 on average and 5.58 of the top 10 slots went to tests and
-    docs; with the penalty, 1.75 and 4.41. Six doc-seeking prompts went
-    2.16 -> 3.00 docs in the top 10, because demoting tests leaves room rather
-    than competing with documentation. Treat those numbers as DIRECTIONAL: the
-    prompts are self-chosen and unlabelled, the corpus is a single Python
-    repo, and a mean over an unbounded rank is dominated by its worst case.
-    The harness is not in the tree, so nobody -- including the author -- can
-    re-run them as written.
+Measured with `tools/rank_eval.py` -- recall@k against hand-labelled
+    relevant files, over 12 prompts on this repo:
 
-    Keyword-only and REQUIRED. It was briefly optional-defaulting-to-False so
+        k        main     with the penalty
+        3        0.250    0.333
+        5        0.542    0.667
+        10       0.667    0.667
+        20       0.667    0.667
+
+    The win is in ORDERING, not retrieval: the same files are reachable
+    either way, they arrive earlier. That matters because per-file character
+    caps mean a file at rank 3 contributes far more content than the same
+    file at rank 9.
+
+    Two earlier metrics -- mean rank of the first source file, and count of
+    tests and docs in the top 10 -- reported a much larger win (4.50 -> 2.00,
+    5.58 -> 4.25) and should not be trusted. Both reward ANY `src/` file
+    landing early regardless of relevance, both penalise a test file that is
+    the correct answer, and both IMPROVE when a file is evicted below
+    `MIN_SCORE_THRESHOLD` rather than merely demoted. They were the reason
+    two other candidate fixes were rejected, and at least one of those
+    rejections was made on a reading those metrics could not support.
+
+        Keyword-only and REQUIRED. It was briefly optional-defaulting-to-False so
     that existing pure-scoring tests would not need editing -- which is a
     production default shaped around test convenience, and the default was the
     buggy behaviour. The call site reads `not prompt_targets_tests(...)`, so a
--- a/src/neo/context_gatherer.py
+++ b/src/neo/context_gatherer.py
@@ -28,6 +28,35 @@ MAX_CHUNKS_PER_FILE = 2    # Cap chunks per file so one large file doesn't domin
 MAX_CHUNK_CENTERS = 20     # Best-scoring lines considered as window centers before merging
 MAX_MERGED_WINDOW_LINES = 200   # Ceiling on a merged window so one file can't eat the budget
 DISCRIMINATIVE_MAX_LINE_FRACTION = 0.25  # A token on >25% of a file's lines is noise in it
+TEST_PENALTY = 0.4         # Multiplier; a test file retains 60% of its score
+
+
+def is_test_path(rel_path: str) -> bool:
+    """Whether `rel_path` looks like a test file."""
+    return (
+        rel_path.startswith("test")
+        or rel_path.startswith("tests" + os.sep)
+        or "/tests/" in rel_path
+        or os.path.basename(rel_path).startswith("test_")
+    )
+
+
+# Word-boundary anchored, which the substring version this replaces was not.
+# `"spec" in prompt` fires on "inspector", "specific", "respective" and
+# "aspect"; measured, "the A2UI inspector shows a stale fact count" was
+# classified as a prompt about testing and every test file kept full score.
+# That was survivable while the flag only softened a cosine boost, and stops
+# being survivable now that it gates the keyword path too.
+#
+# `\btest` is a prefix rather than a whole word so "testing" and "tests"
+# match, while "latest" does not -- there is no boundary before its `test`.
+_TEST_PROMPT_RE = re.compile(r"\btest|\bpytest\b|\bspecs?\b", re.IGNORECASE)
+
+
+def prompt_targets_tests(prompt: str) -> bool:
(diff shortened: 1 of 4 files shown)


JLiotta-25:
## Correction: the metrics in the description were not fit to judge this

I built a proper instrument (`tools/rank_eval.py`, added in 34e3f0b) and re-measured. **The change is still worth landing, but its benefit is narrower and differently shaped than the description claims.**

### What the harness says

| k | main | this branch |
|---|---|---|
| recall@3 | 0.250 | **0.333** |
| recall@5 | 0.542 | **0.667** |
| recall@10 | 0.667 | 0.667 |
| recall@20 | 0.667 | 0.667 |

Flat at k=10 and above. The gain is in **ordering, not retrieval** — the same files were reachable either way, they now arrive earlier. That matters for a concrete reason the description never gave: per-file character caps mean a file at rank 3 contributes far more content than the same file at rank 9.

### Why the original numbers were wrong

"Mean rank of the first source file" (4.50 → 2.00) and "tests+docs in top 10" (5.58 → 4.25) overstate it, and are wrong in three ways:

1. They reward **any** `src/` file landing early, regardless of relevance.
2. They **penalise a test file that is the correct answer**.
3. They **improve when a file is evicted** below `MIN_SCORE_THRESHOLD` rather than merely demoted — so the instrument was blind to the worst thing the change could do.

Measured live: for "add a new CLI subcommand for exporting facts", a candidate change surfaced `tests/test_subcommands.py` and `src/neo/prompt/cli.py` — both relevant — and scored **worse** than a baseline whose top hits were `.pytest_cache/README.md`, `docs/mistakes.md` and `construct/README.md`.

### A third fix was built and rejected on the new harness

Both fixes rejected in the description attacked *which tokens exist*. Neither addressed the mechanism: a one-character token is **substring-matched against a path**. `"a" in "src/neo/subcommands.py"` is meaningless; `"db" in "src/db_helper.py"` is not — and neither frequency nor length distinguishes them.

Requiring short tokens to match a whole filename token (leaving 3+ char tokens on substring matching) does address it. Measured **0.333 / 0.583 / 0.583 / 0.625** against this branch's **0.333 / 0.667 / 0.667 / 0.667** — equal at k=3, worse everywhere else. Not shipped.

So the underlying defect stands, now with a documented reason and a cheap way to judge the next attempt.

Harness limits are written into the tool: 12 prompts, one repository, labels by one person, and recall says nothing about what else got in. Better than a rank mean, not good.

JLiotta-25:
Subsumed by the content-BM25 work, which carries all three of this branch's commits (cherry-picked) plus the change that makes them matter.

**Why this closes rather than merges:** #188 tuned the filename-and-byte-count scorer. That scorer is deleted — file selection now ranks by BM25 over file *content*, because it turned out selection never read the file at all. Merging this separately would land a tuning of code that no longer exists.

**The test-file demotion itself survives, and is worth more than it was here.** Re-validated against the new baseline (stage 2 of the plan), real-pipeline A/B on identical code:

| repo | stage 1 only | + demotion | Δ MRR |
|---|---|---|---|
| neo | 0.564 | **0.718** | +0.154 |
| car | 0.643 | **0.646** | +0.003 |
| quip | 0.572 | **0.634** | +0.062 |

The reason it matters more under BM25: a test file is a lexical *superset* of its subject — same identifiers plus scaffolding — so BM25 ranks it at least as high every time. Under the old scorer that was one signal among several; under content ranking it is the ranking.

Two things from this PR were re-expressed rather than carried over verbatim:

- **Filename overlap drops 0.6 → 0.15 per hit.** It is a substring test against the whole path, so short tokens match by coincidence (`a` is inside 49 of 85 basenames here). At 0.6 that noise competed with real signal; at 0.15 it breaks ties.
- **This PR's tests needed rewriting, not patching.** They compared two files whose *only* signal was their names — a premise content ranking removes. They now hold content equal and let the penalty be the only difference, which is the scenario the penalty exists for.

Also worth recording: the correction posted on this PR stands. The metrics originally used to justify it (mean rank of first source file; tests+docs in top 10) reward any `src/` file landing early regardless of relevance, penalise a test file that is the correct answer, and *improve* when a file is evicted from context entirely. `tools/rank_eval.py` replaced them and is what every number above was measured with.

Tracking: #193. Superseding branch: `fix/file-selection-content-bm25`.

PR #190, opened 2026-08-10
feat(index): give ruby and php the edge queries they never had

JLiotta-25:
## Description

Ruby declared **no edge queries at all** — the only language in `QUERIES` absent from `EDGE_QUERIES`. Every ruby file in every index contributed nodes to the graph and never a single relationship: no `require`, no superclass, no mixin. PHP declared imports only, so `class Impl extends BaseImpl implements Greeter` produced nothing where the java, c_sharp and typescript equivalents each produce edges.

Both gaps were found while writing the per-query yield tests in #187 and pinned there as failing-when-fixed assertions. Both pins have now fired.

## Type of Change

## Related Issue

Follow-on from #158. **Stacked on #187** — see Additional Notes.

### Ruby: the method-call problem, and the predicate that solves it

The existing code carried a comment explaining the omission: *"Ruby imports are method calls (`require`, `require_relative`) rather than syntactic forms — adding them would need either a method-call pattern match or a runtime-style heuristic."* The method-call pattern match is what this uses.

Because `require` is an ordinary call, the pattern matches **any** call taking a string — only the `#match?` predicate separates an import from `puts 'hello'`. That predicate is load-bearing, so its enforcement is a test: if a future tree-sitter stopped applying predicates in `QueryCursor.captures`, the query would keep compiling and start emitting an import edge per string-argument call. Silent, and in the noisiest possible direction.

### The first cut dropped every namespaced supertype, silently

This is the part worth reading. `(base_clause (name) @base)` requires `name` as a **direct** child; the real shape is `base_clause → qualified_name → name`, and Ruby's is `superclass → scope_resolution → constant`. Measured against realistic source, the first version produced:

```
class B extends \Foo\Bar implements \ArrayAccess, Countable {}
  → ONE edge:  B implements Countable

class ApplicationRecord < ActiveRecord::Base
  include ActiveSupport::Concern
  → ONE edge:  ApplicationRecord implements Plain
```

`ActiveRecord::Base` is the single most common inheritance statement in the Ruby ecosystem and produced **nothing**. This is silent *partial* extraction, which is worse than a total miss: the consumer sees a plausible answer with no signal that most of it is absent.

It is also precisely what the `_CS_BASE_TYPES` comment **150 lines above the new code** warns about — and I quoted the other half of that comment in my own rationale while writing the narrow arm.

Note: do **not** copy the C# spelling `(qualified_name name: (name) @x)`. PHP's `qualified_name` has no `name:` field; that pattern fails to compile.

### Two more gaps and one consistency decision

- **PHP 8.1 enums** carry `class_interface_clause` and were uncovered.
- **PHP `use SomeTrait;`** is the same construct Ruby's mixins are. Shipping two languages in one commit with opposite stances on it would not have survived the next maintainer, so traits map to `implements`, as `include`/`extend`/`prepend` do.
- The PHP alternation is now **generated** from `_PHP_BASE_TYPES` across declaration kinds, per the `_CS_INHERITANCE` precedent — it had already grown to three declaration kinds across two queries, hand-copied, which is the exact state that comment exists to prevent.

Query **names** are load-bearing and easy to get wrong: `extract_edges` dispatches on exactly `imports` / `import_from` / `inheritance` / `implements` and silently ignores anything else, so a well-formed query under any other name is a no-op.

### The fixtures were the root cause

They were minimal and written by the same person as the queries, so they agreed with them by construction — 5/5 and 6/6 mutation kills measured the *fixture*, not the grammar. They are now framework-shaped: Rails (`< ActiveRecord::Base`, `include ActiveSupport::Concern`) and PSR/Laravel (`extends \Foo\BaseImpl implements \ArrayAccess`, an 8.1 enum, a trait `use`).

The module docstring's "wait until a real ruby repository is available" limit was also wrong, and that mistaken framing is what let the narrow arms ship: no ruby or php **corpus** exists locally, but the **grammars** are installed. The limit is now stated accurately.

## How Has This Been Tested?

- [x] `2132 passed, 8 skipped`; ruff clean
- [x] Every one of the five new queries mutation-tested individually — 5/5 caught
- [x] Plus four more: stripping the `#match?` predicate, re-narrowing the PHP arms, re-narrowing the Ruby arms, dropping `enum_declaration`
- [x] All 11 expected edges extract from framework-shaped source, verified against the installed grammars

## Checklist

## Additional Notes

**Stacked on #187, deliberately.** `tests/test_language_query_yield.py` does not exist on `main` — it arrives across three commits of that branch, and the two pins this PR must flip live only there. Rebasing onto `main` would mean either dropping the pins (losing the point) or carrying a duplicate of a 290-line file. **Merge #187 first.**

**Standing limit:** the fixtures are hand-written framework snippets, not a corpus. A green run means the queries work on the shapes we know about, not that they are complete — PHP group `use App\{A, B};` is a known pre-existing gap.

### DIFF
--- a/src/neo/index/language_parser.py
+++ b/src/neo/index/language_parser.py
@@ -303,6 +303,62 @@ _CS_INHERITANCE = "[\n" + "\n".join(
 ) + "\n] @class_def"
 
 # Edge queries for relationship extraction (imports, inheritance)
+# PHP supertypes reach the query in two shapes and BOTH must be covered:
+#
+#     implements Countable          -> (name)
+#     extends \Foo\Bar             -> (qualified_name (name))   (everyday PSR)
+#
+# The narrow arm alone compiles perfectly and drops every namespaced
+# supertype without a word -- `class B extends \Foo\Bar implements
+# \ArrayAccess, Countable` yielded exactly one edge, `B implements
+# Countable`, which is worse than a total miss because the consumer has no
+# signal anything is absent. This is the same failure the `_CS_BASE_TYPES`
+# comment above documents, reproduced in a new language 150 lines below it.
+#
+# Do NOT copy the C# spelling `(qualified_name name: (name) @x)` -- PHP's
+# `qualified_name` has no `name:` field and that pattern fails to compile.
+# The unfielded direct child is the last segment, which matches C#'s
+# rightmost-segment convention.
+_PHP_BASE_TYPES = "[(name) @{cap} (qualified_name (name) @{cap})]"
+
+# Generated across declaration kinds rather than written out per kind, for
+# the reason `_CS_INHERITANCE` is generated: a hand-maintained duplicate
+# invites the next widening to be applied to some of the arms and not the
+# rest. PHP 8.1 enums carry `class_interface_clause` too, and were missed by
+# the hand-written version.
+_PHP_EXTENDS_KINDS = ("class_declaration", "interface_declaration")
+_PHP_IMPLEMENTS_KINDS = ("class_declaration", "enum_declaration")
+
+
+def _php_edge_queries() -> dict:
+    base = _PHP_BASE_TYPES.format(cap="base")
+    iface = _PHP_BASE_TYPES.format(cap="interface")
+    extends = " ".join(
+        f"({kind} name: (name) @class_name (base_clause {base}))"
+        for kind in _PHP_EXTENDS_KINDS
+    )
+    implements = " ".join(
+        f"({kind} name: (name) @class_name (class_interface_clause {iface}))"
+        for kind in _PHP_IMPLEMENTS_KINDS
+    )
+    # `use SomeTrait;` inside a class body. Mapped to `implements` for parity
+    # with Ruby's mixins in this same change: both mean "this type gains that
+    # module's surface", and two languages in one commit taking opposite
+    # stances on the same construct would not survive the next maintainer.
+    traits = (
+        f"(class_declaration name: (name) @class_name "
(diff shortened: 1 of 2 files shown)


JLiotta-25:
**CI cannot run on this PR as stacked.** `.github/workflows/ci.yml` triggers only on `pull_request: branches: [main]`, and this targets `fix/dry-run-shows-what-it-claims`. GitHub will report "no checks reported" indefinitely — that is not a pending run, and it must not be read as green.

Merge order:
1. Merge #187.
2. Retarget this PR to `main` (`gh pr edit 190 --base main`). CI fires then.
3. Merge once green.

Verified locally in the meantime: `2132 passed, 8 skipped`, ruff clean, on Python 3.13.9.

kkeeling:
**Heads-up: this PR reads MERGED, but its changes are not on `main`.** It was squash-merged into its base branch `fix/dry-run-shows-what-it-claims` (#187) as `54a24df`, and that branch was subsequently rebased onto `main` from the tip *before* that merge (`64e1ee1`) as part of Goal 2's queue landing. The force-push took `54a24df` with it, and #187 then squash-merged carrying its own five commits and none of yours.

Verified against `main` rather than assumed: `src/neo/index/language_parser.py` on `main` has one occurrence of `'ruby'` where this branch has two, and is 111 lines short.

This PR cannot be reopened (`already merged`) or retargeted (`cannot change the base branch of a closed pull request`), so the work is re-landed byte-identically, with your authorship preserved on both commits, in **#204**. Nothing here is lost or rewritten.

PR #192, opened 2026-08-10
tools: add a retrieval evaluation harness for file selection

JLiotta-25:
## Description

Adds `tools/rank_eval.py` — recall@k and MRR for file selection, with eval sets mined from git history. **No behaviour change.**

This is the harness the diagnosis in #193 was measured with, and the one its acceptance criteria are written against. It is useful independently of when that fix lands.

## Type of Change

- [x] Code cleanup/refactoring (tooling; no runtime behaviour touched)

## Related Issue

Measurement for #193. The plan itself lives outside the repo at `~/git/working/2026-08-10-neo-file-selection-plan.md` — a plan for work that has not started does not belong in the tree, where it reads as documentation of behaviour that exists.

## Motivation and Context

Three separate metrics were used to justify file-ranking changes this week:

- mean rank of the first source file
- count of tests and docs in the top 10
- a 12-prompt hand-labelled set

All three are wrong in the same three ways. They reward **any** `src/` file landing early regardless of relevance; they **penalise a test file that is the correct answer**; and they **improve when a file is evicted** below `MIN_SCORE_THRESHOLD` rather than merely demoted — so the instrument was blind to the worst outcome a change could produce.

Measured live: for *"add a new CLI subcommand for exporting facts"*, a candidate change surfaced `tests/test_subcommands.py` and `src/neo/prompt/cli.py` — both relevant — and scored **worse** than a baseline whose top hits were `.pytest_cache/README.md`, `docs/mistakes.md` and `construct/README.md`.

## What it does

```
python tools/rank_eval.py --build-from-git <repo> > cases.json
python tools/rank_eval.py --eval <repo> cases.json
```

Labels come from git history — commit subject as the query, changed non-test files as ground truth, keeping commits that touch 1–3 files. That is the standard construction in the bug-localization literature, and it removes the two worst properties of a hand-written set: prompts chosen by the person evaluating the change, and labels that are that person's opinion. 506 cases across three repos.

It reports four strategies side by side (current scorer, content BM25, dense, RRF hybrid) at k = 1/3/5/10/20 plus MRR, so a change is judged on retrieval quality rather than on a proxy.

## How Has This Been Tested?

- [x] `2054 passed, 8 skipped`; ruff clean — nothing in `src/` changed
- [x] Reproduces every number in #193, on neo, car and quip
- [x] Runs against foreign repos with no ProjectIndex present

**Test Configuration**: Python 3.13.9 · macOS 26.5.2 · neo 0.44.0

## Checklist

[...]

PR #194, opened 2026-08-11
fix(gatherer): select files by content, not by filename and byte count

### LINKED ISSUE #193: File selection ranks files without reading them; size penalty has the sign backwards
## The defect

`score_candidate` (`src/neo/context_gatherer.py:338`) takes `(rel_path, size, prompt_tokens, git_recent, entry_points)`. **There is no content parameter.** Files are ranked by path string and byte count; content is first read at `context_gatherer.py:848`, *after* selection.

With no content signal, relevance is inferred from a filename, so the scorer is a stack of hand-tuned additive bonuses — and the dominant term is anti-correlated with relevance.

## Size penalty

`score -= 0.01 * size_kb` once over 10 KB, **uncapped**, against a realistic positive signal of +0.6 to +2.1. A file with one keyword hit becomes unrankable above 60 KB.

Ground-truth files measured 31–177 KB; corpus median is 10 KB. Central files are large *because* they are central.

Measured, for *"fix the fact store supersession threshold"*:

```
src/neo/memory/store.py   keyword +0.60   depth -0.15   size -1.62   ->  0.000
                                                        rank 200 of 284
```

This was already diagnosed in-tree — the comment at `context_gatherer.py:487` says *"they're large because they're central, and the old 0.002 multiplier was pushing THE relevant file (93KB engine.py) below threshold"* — and patched with a seven-name stem whitelist:

| file | KB | size penalty | whitelisted |
|---|---|---|---|
| `src/neo/engine.py` | 177 | **-0.13** | yes |
| `src/neo/memory/store.py` | 162 | **-1.62** | no |

A 12x disparity between two files of near-identical size, decided by whether someone thought of the name.

BugLocator/rVSM (Zhou et al., ICSE 2012) ranks **larger files higher** for this exact task, because larger files are empirically more likely to contain the defect. The sign is backwards, not just the magnitude.

## The semantic path has the same bias independently

`MAX_CHUNKS_PER_REPO = 1000` with round-robin allocation gives every file ~6 chunks:

| file | functions | indexed | coverage |
|---|---|---|---|
| `src/neo/memory/store.py` | 82 | 6 | **7%** |
| `src/neo/engine.py` | 95 | 6 | **6%** |
| `src/neo/text_budget.py` | 4 | 4 | 100% |

`src/` alone produces 1,344 chunks against the cap. The index is neither stale nor missing — it is truncated.

## Impact

Measured over 506 cases mined from git history (commit subject = query, changed non-test files = ground truth), across three repos:

| repo | cases | R@10 current | R@10 with content BM25 | MRR |
|---|---|---|---|---|
| neo | 207 | 0.223 | **0.697** | 0.143 -> 0.565 |
| car | 75 | 0.382 | **0.969** | 0.296 -> 0.621 |
| quip | 224 | 0.137 | **0.762** | 0.107 -> 0.544 |

On quip the correct file reaches the top 10 **14% of the time**.

## Fix

BM25 over file content with code-aware tokenization. Neo already ships the BM25 (`src/neo/memory/bm25.py`) and already runs hybrid dense+sparse for *fact* retrieval; file selection uses none of it.

**Plan and staged work:** `~/git/working/2026-08-10-neo-file-selection-plan.md`
**Goal prompt:** `~/git/working/2026-08-10-neo-file-selection-goal-prompt.md`
**Harness:** #192 adds `tools/rank_eval.py`, which reproduces every number above.

Assigned so nobody else picks it up.

JLiotta-25:
## Description

**File selection ranked files without reading them.** `score_candidate` took `(rel_path, size, prompt_tokens, git_recent, entry_points)` — no content parameter. The file was first opened inside `gather_context`, *after* selection, only to chunk what had already been chosen.

Every other defect followed from that. With no content signal, relevance had to come from a filename and a byte count, and the dominant term was `score -= 0.01 * size_kb` — once over 10 KB, **uncapped**, against a realistic positive signal of +0.6 to +2.1. So a file with one keyword hit was unrankable above 60 KB.

Measured, for *"fix the fact store supersession threshold"*:

```
src/neo/memory/store.py   keyword +0.60  depth -0.15  size -1.62  =  0.000
                                                      rank 200 of 284
```

The most relevant file in the repository, unrankable, because it is 162 KB. Ground truth ran 31–177 KB against a corpus median of 10 KB — central files are large *because* they are central.

This had been diagnosed once already. The comment above the penalty read *"they're large because they're central, and the old 0.002 multiplier was pushing THE relevant file (93KB engine.py) below threshold"* — and the fix was a seven-name stem whitelist that rescued `engine.py` (−0.13) and left `store.py` (−1.62). A **12× disparity** between two files of near-identical size, decided by whether someone had thought of the name.

**The literature says the sign was wrong, not the magnitude.** This task is IR-based bug localization; BugLocator's rVSM (Zhou et al., ICSE 2012) adds a logistic length function that ranks *larger* files *higher*, because larger files are empirically more likely to contain the defect. BM25's `b` handles the same concern with bounded, corpus-derived length normalization.

## Type of Change

- [x] Performance improvement

## Related Issue

Fixes #193.

## Motivation and Context

Neo already shipped the BM25 (`neo.memory.bm25`, Lucene-style with IDF smoothing) and already ran hybrid dense+sparse fusion for **fact** retrieval. File selection used none of it.

Now: BM25 over file content with code-aware tokenization — identifiers split on camelCase and separators, emitting both the whole identifier and its parts, so a prompt saying "user by id" reaches `getUserById`. Score max-normalized per query and applied as the dominant term, at a named `CONTENT_WEIGHT = 3.0`. The size penalty and the `main_impl_stems` whitelist are deleted; filename overlap drops from 0.6 to 0.15 per hit and becomes a tie-breaker; `EXPLICIT_PATH_BOOST` is untouched.

### Measured, end-to-end through the real CLI

Cases mined from git history — commit subject as query, changed non-test files as ground truth. `PYTHONPATH` pinned per tree, at the shipped `CONTENT_WEIGHT = 3.0`, 50 cases per repo, zero failed cases:

| repo | R@1 | R@10 | MRR | top-10 hit rate |
|---|---|---|---|---|
| neo | 0.081 → **0.351** | 0.301 → **0.742** | 0.304 → **0.771** | 0.58 → **0.98** |
| car | 0.047 → **0.185** | 0.180 → **0.472** | 0.162 → **0.425** | 0.24 → **0.62** |
| quip | 0.057 → **0.321** | 0.174 → **0.696** | 0.158 → **0.643** | 0.30 → **0.86** |

Reproduce with `tools/rank_mine_eval.py --repo <repo> --tree <tree>`, which ships in this PR.

**Measured with the recency signal off, which is the harness default.** `get_git_recent_files` reads `git status --porcelain` plus the last 50 commits and stores *paths*, so a case mined below the window whose truth file was touched again inside it is handed its own answer key — as is every file dirty in the tree you measure from. `neo --dry-run --no-git` gates that signal and nothing else, leaving `_history_boost` and the rest of the re-rank live, which is what `tools/rank_eval.py` had been doing all along. Running `--with-git` instead reports **48 of 50 neo cases contaminated** and moves every figure by ≤0.03: the leak is real, and it was never what carried the result.

> **These supersede the table this description carried until now** (neo 0.078 → 0.603 / 0.082 → 0.655, and so on). Not a correction of arithmetic — a different instrument. The original numbers came from a harness that was never committed and no longer exists to re-run, which is the defect this revision fixes; see *Corrections after review*. Same direction, different magnitudes, and only these are reproducible.

### Two measurement bugs, both of which produced confident wrong numbers

Worth more than the result, because both are easy to hit again and both are now documented in CLAUDE.md.

1. **The venv installs neo EDITABLE against `src/`.** Running `.venv/bin/python -m neo.cli` from a git worktree of another commit executes *this* tree's code against *that* tree's files. My "main baseline" was branch code. It made main look like MRR 0.613 against a real figure of 0.082 — a 7× error, flattering to the change under test. (Both are superseded-generation figures, quoted for the size of the error, not as measurements; this harness puts main's neo MRR at 0.304.) Fixed with `PYTHONPATH=<tree>/src`.
2. **Calling `score_candidate` directly measures only the first-pass ranking.** `gather_context` then re-ranks with `pi_boost + hist_boost + _symbol_score` and applies an adaptive limit and byte budget. A first-pass harness overstated R@10 by 0.14 against the real CLI.

Both are now written into `tools/rank_mine_eval.py`'s docstring, and the tool closes them structurally: it shells out to `neo --dry-run` rather than replicating the pipeline, and it absolutizes `--tree` and then **refuses to run unless that tree is the one `import neo` actually resolves to**. Mandatory is not the same as effective — review demonstrated that a typo'd `--tree` falls through the editable install's `.pth` to the working checkout, so both arms measure the same code and the run still reports `failed_cases: 0`. That is trap 1 returning inside the tool written to prevent it. The guard then had to be fixed twice for the same reason it exists — checking one directory while the run measured another. First, a relative `--tree` resolved against the harness's directory while the child ran with `cwd=repo`, so `--repo ../car --tree .` passed and Python discarded the nonexistent `PYTHONPATH` entry in silence. Second, the probe itself ran from the wrong place: cwd lands at `sys.path[0]` *ahead of* `PYTHONPATH`, so a repo with a top-level `neo/` would shadow the named tree in the measurement while a probe run from elsewhere saw nothing wrong. The probe now runs from `cwd=repo`, exactly as the measurement does. No repo involved here has such a directory, so no published number was affected — but a guard is the one place "nobody has hit it yet" is not an argument. I reported wrong numbers in both directions before arriving at these. Any in-process replica must be validated against `--dry-run` before a sweep is run on it.

### Also in this branch

- **PR #188's test-file demotion**, cherry-picked and reconciled — it earns its place and is worth *more* under BM25, because a test file is a lexical superset of its subject (same identifiers plus scaffolding). The figures that established this came off a superseded harness generation and are deliberately not restated here, per `score_candidate`'s own docstring; the direction held on every repo and is carried by the table above. #188 is closed as subsumed.
- **Proportional chunk allocation** (`_cap_chunks`, reusing the existing `_allocate_slots`). Round-robin gave every file the same *count* regardless of size — `store.py` 6/82 symbols = 7% coverage, `text_budget.py` 4/4 = 100%. Now equal coverage *fraction*: store.py 6 → 19 chunks.

## Corrections after review

Two rounds of review. The first found a blocker that had been shipped **and measured** without being noticed, plus a conclusion already written into CLAUDE.md that was wrong — all fixed in `7d00723`, full write-up in [this comment](https://github.com/Parslee-ai/neo/pull/194#issuecomment-5250285666). The second found that the evidence for the whole change was unreproducible.

**The harness that produced the headline table was never committed**, and CLAUDE.md named `tools/rank_eval.py` as its source — a different instrument entirely: 12 hand-labelled prompts, this repo only, recall@k, no MRR, no repo argument. A false claim about a file in the same commit, and the same paragraph credited it with two docstring warnings it does not contain. Nobody, including me, could re-run the numbers this PR rests on.

`tools/rank_mine_eval.py` now ships as that harness, and re-measurement replaced the table above. Its own first draft then failed review twice over, which is worth recording because both failures are the exact defect this branch exists to remove:

- **It claimed `--skip-recent` removed recency contamination. It does not.** The scorer's recency set holds paths rather than commits, so a truth file touched again inside the window is back in it, and `get_git_recent_files` also reads `git status --porcelain` — so measuring a branch from the tree you are editing leaks regardless of history. My first sweep ran against a dirty tree. Worse, I answered this by inventing a `--drop-contaminated` sampling filter when **`neo --dry-run` has taken `--no-git` all along, and the sibling harness in the same directory was already passing it** — a bespoke workaround for a solved problem, sitting one file away. That flag gates `git_recent` and nothing else, so it removes the leak structurally instead of by discarding data. It is now the default; `--with-git` opts back in and reports contamination counts.
- **Its parser failed open.** A timeout, a non-zero exit or a renamed heading returned an empty ranking, which scored as perfect-zero retrieval and averaged into the table as if it were evidence about ranking quality. Failed cases are now dropped and counted, never scored. The published sweep had zero.
- **Fixing that at the header was not enough.** A per-*line* format change keeps the banner and parses to nothing, so every case returns empty and the run reports a confident `R@10 0.000` with no failures — the same defect back through a different door, and there is a live trigger today (the CLI's second dry-run printer emits absolute paths, which parse cleanly and match no repo-relative ground truth). Marker found but nothing parsed anywhere is now a hard failure.
- **Reverts were being mined as cases.** `revert:` and `wip:` are themselves conventional-commit types, so stripping the prefix consumed the very word that should have skipped them and the remainder read as ordinary work. A revert's changed files are a removal, which is not ground truth for its own subject. Caught by writing the test against the predicate rather than against a composition performed in the test body; the case sets turned out identical on all three repos, so the table above is unaffected — verified rather than assumed.

- **The content term was applied twice.** `score += 3.0 * content_relevance` sat at two adjacent lines, so the effective weight was **6.0** while every docstring, comment and table said 3.0 — introduced by the cherry-pick that merged #188's scorer with this one. Now a named `CONTENT_WEIGHT` applied once, with a structural test that it is applied exactly once. Re-measurement showed 3.0 beats 6.0 — but that is luck, not process. **The honest statement is that every number previously reported on this PR was produced at a weight nobody chose.**
- **"The re-rank is redundant" was wrong.** Disabling `pi_boost` + `_symbol_score` + `hist_boost` on the real CLI takes **R@1 from 0.344 to 0.044** and MRR from 0.646 to 0.261 — an ablation run on the superseded harness generation, so read it against itself, not against the table above. It is load-bearing. The opposite conclusion came from a sweep at k=10 — where every configuration on this repo is flat — through an in-process replica omitting the byte budget and adaptive limit, i.e. exactly the stages that make the re-rank matter. Both mistakes are now warned about in `tools/rank_mine_eval.py`'s docstring — an earlier revision of this description attributed those warnings to `rank_eval.py`, which carries neither. The surviving half is real and kept: `hist_boost` contributes nothing measurable.
- **Binary files were entering the BM25 corpus.** `errors="ignore"` turned PDFs into junk tokens, and they set `avgdl` — so length normalization depended on how many papers happened to sit in the repo. The same defect this change exists to remove, re-entering through corpus statistics. NUL-byte probe added. Re-measured through the real candidate path for this description: **8 of 289 candidates (2.8%) carrying 27.5% of tokens**, `avgdl` 2760 → 2059. The exact share is not stable — four of the eight are `.ruff_cache` blobs, so it depends on whether ruff has run since the last clean, and an earlier reading on a cleaner tree recorded 5 PDFs at 1.7%/32.9%. The stable shape is a low single-digit share of documents carrying roughly a third of corpus tokens.
- **`MAX_INDEXED_BYTES` → `MAX_INDEXED_CHARS`.** `TextIOWrapper.read(n)` counts code points; CLAUDE.md already records this exact lie in `_render_context_files`.
- **Four constants pinned by value** (`CONTENT_WEIGHT`, `MAX_INDEXED_CHARS`, `PATH_TOKEN_WEIGHT`, the normalize ceiling). The test that should have caught the doubling named 3.0 in its docstring and asserted an inequality satisfied by anything above 1.76 — naming a number in prose while asserting a direction is how a constant goes unpinned and looks pinned. To be exact about what the pins buy: none of them would have caught this bug, because the constant *was* 3.0 and the doubling was at the call site. Only `test_content_weight_is_applied_exactly_once` catches that. The pins close the adjacent hole.
- The chunk-cap console line no longer promises a floor that `_allocate_slots` does not always give, and three conflicting measurement tables were reconciled to one with the harness generation named (`car` had appeared as both 0.969 and 0.507 for the same quantity).

## Negative results — recorded so they are not blindly retried

| tried | result |
|---|---|
| RRF fusion with the dense channel | **Provisional — must be retried, not filed away.** Measured worse than BM25 alone (0.596 best-weighted vs 0.693 — superseded-generation figures, unstamped when first recorded, and **not** comparable with the headline table above), but the result is triply compromised: those two numbers come off a harness that no longer exists, it was taken at k=10 (the blind cutoff that produced the re-rank error above), and the mechanism it blamed — dense's narrow ~25-file recall against BM25's ~180, compounded by chunk starvation — **was fixed two commits later in this same branch**. Re-measure at k=3/k=5 against the current tree before relying on it either way. |
| Evidence/saturation term on the BM25 score | +0.010 MRR (superseded generation) — noise. Every eval prompt is an on-topic commit subject, so the harness cannot see the prompt class it serves. Documented, not shipped. |
| Four filename-tuning fixes (document frequency, `{3,}` length floor, whole-token matching, cutting the docs bonus) | All measured and rejected in earlier work. They tune the *filename*; the filename is not the evidence. |

The re-rank weight sweep is **not** in this table. It concluded the re-rank was redundant and was refuted — see *Corrections after review*.

## Deferred

**#195 — BM25 is O(N) in memory and query time** (~3 s and ~100 MB per 1,000 files; ~15 s and ~500 MB at 5,000). The fix is a precomputed per-doc `Counter` plus an inverted index in `src/neo/memory/bm25.py`, which has other callers (fact retrieval), so it is not this PR's job. `gather_context` is CLI-only, so the observer daemon's documented RSS behaviour is unaffected. Filed with measurements and blast radius stated.

## How Has This Been Tested?

- [x] `2162 passed, 8 skipped`; ruff clean
- [x] End-to-end on three repos via the real CLI with `PYTHONPATH` pinned per tree, zero failed cases
- [x] The harness itself is tested (`tests/test_rank_mine_eval.py`, 41 tests): the recall/hit-rate/MRR arithmetic, the fail-closed paths for non-zero exit, missing marker and timeout, chunk de-duplication, and that `--no-git` is passed unless asked otherwise. A defect here does not crash — it publishes a wrong number with a confident table around it, which is this branch's whole subject
- [x] Contamination audited after review: `_history_boost` returns 0 entries for dev-run prompts; the main worktree was given its own ProjectIndex and scores **identically** (0.082 — superseded generation, from the audit run at the time), so the comparison is fair; `iter_paths` returns a list, so `build_index` cannot exhaust the candidate iterable
- [x] Mutation-tested: 4 mutants on `file_retrieval`, 4 on the scorer, 2 on `_cap_chunks` — all caught, after two gaps were found and closed, and the set re-run after the weight correction
- [x] Cost measured on this repo at 279 indexed files: **0.48 s** to walk, read and build the index (0.07 s of it I/O), **68 ms** for one query. No cache; the gatherer already reads every file it selects. #195 is the authority on how this scales (~3 s and ~100 MB per 1,000 files) — earlier revisions of this description quoted 0.14–0.16 s build and 12–14 ms query, which do not reproduce and conflicted with #195 by roughly 8×
- [x] Ablation (superseded generation — quoted for relative scale, not as current measurements): git recency +0.021 R@10, filename overlap +0.009, docs bonus −0.002 on code prompts but **+1.0 docs-in-top-10 on doc-seeking prompts**, so it stays

## Checklist

## Additional Notes

**Known limits, stated rather than fixed.** Max-normalization preserves order but discards evidence — the top file gets the full ceiling whether its raw BM25 score was 20 or 4. Total abstention works (a query matching nothing yields no boost), but a weak best match is promoted as confidently as a strong one; measured separation is ~2× (on-topic 14.8–20.2, off-topic 6.3–7.5). Ground truth is what a commit *changed*, a subset of what a developer needed to *read*, and commit subjects are terser and better-formed than real prompts.

**The docs bonus and the saturation term are mirror decisions.** An eval that cannot see doc-seeking prompts was not allowed to justify *deleting* a feature serving them; an eval that cannot see off-topic prompts was not allowed to justify *adding* one. Neither was decided on argument.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -541,9 +541,12 @@
   (1) The venv installs neo EDITABLE against `src/`, so running
   `.venv/bin/python -m neo.cli` from a git worktree of another commit executes THIS
   tree's code against that tree's files — a "baseline" run that is not the baseline.
-  Every A/B needs `PYTHONPATH=<that tree>/src`. Measured on the superseded harness
-  generation: it made main look like MRR 0.613 against a real figure of 0.082 (the
-  current harness puts main at 0.294 — different instrument, same trap). (2) Calling `score_candidate` directly
+  Every A/B needs `PYTHONPATH=<that tree>/src`, and `rank_mine_eval` now REFUSES to
+  run unless the tree it was handed is the tree `import neo` actually resolves to —
+  mandatory is not the same as effective, since the editable `.pth` silently catches
+  a typo'd path and measures the working checkout twice. Measured on the superseded
+  harness generation: it made main look like MRR 0.613 against a real figure of
+  0.082 (the current harness puts main at 0.304 — different instrument, same trap). (2) Calling `score_candidate` directly
   measures the FIRST-PASS ranking only; `gather_context` then re-ranks with
   `pi_boost + hist_boost + _symbol_score` and applies an adaptive limit and a byte
   budget. A first-pass harness overstated R@10 by 0.14 against the real CLI. Validate
@@ -666,7 +669,9 @@
   was never what carried the result.
   **The re-rank is LOAD-BEARING, not redundant** (`pi_boost` + `_symbol_score` +
   `hist_boost`, applied after the first-pass score). Disabling it on the real CLI
-  takes R@1 from 0.344 to **0.044** and MRR from 0.646 to **0.261**. An earlier
+  takes R@1 from 0.344 to **0.044** and MRR from 0.646 to **0.261** — a
+  SUPERSEDED-generation ablation (pre-`--no-git`, uncommitted harness), so read
+  those four numbers against each other and never against the table above. An earlier
   version of this note claimed the opposite, from a weight sweep that landed
   everywhere in 0.66–0.68 — measured at k=10, where every configuration is flat, and
   through an in-process replica that omits the byte budget and adaptive limit, i.e.
@@ -676,7 +681,8 @@
   contributes nothing measurable (identical results with it disabled) while
   `_symbol_score` carries most of the effect.
   RRF fusion with the dense channel LOSES to BM25 alone (0.596 best-weighted vs
-  0.693) — dense returns ~25 files against BM25's ~180 and is half as accurate. Treat
+  0.693, also superseded-generation and unstamped when first recorded) — dense
+  returns ~25 files against BM25's ~180 and is half as accurate. Treat
   that as provisional: it was measured at k=10 before the same cutoff problem was
   understood, and the docstring deferring it to a chunk-allocation fix is stale
   because that fix landed in the same branch. Re-measure at k=3/k=5 before relying on
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -644,8 +644,8 @@
   this exact task, and BM25's `b` handles the concern with bounded, corpus-derived
   length normalization. Measured end-to-end over cases mined from git history
   (commit subject = query, changed non-test files = ground truth), R@10 / MRR:
-  neo 0.275→0.732 / 0.294→0.787, car 0.192→0.502 / 0.160→0.398, quip 0.175→0.720 /
-  0.166→0.658, at `CONTENT_WEIGHT = 3.0`. **`tools/rank_mine_eval.py` is the
+  neo 0.301→0.742 / 0.304→0.771, car 0.180→0.472 / 0.162→0.425, quip 0.174→0.696 /
+  0.158→0.643, at `CONTENT_WEIGHT = 3.0`. **`tools/rank_mine_eval.py` is the
   harness** — not `tools/rank_eval.py`, which is a different instrument (12
   hand-labelled prompts, this repo, recall@k, no MRR) and was named here in error
   while the real one went uncommitted, leaving the figures unreproducible. Quoting
@@ -653,14 +653,17 @@
   at 0.969 and 0.507; keep the generation attached. An earlier generation of these
   same figures read neo 0.078→0.603 / 0.082→0.655 — superseded, and not comparable,
   because the harness that produced it no longer exists to re-run.
-  **The absolute values are upper bounds, and the tool says so per run.** The
-  scorer's recency signal holds PATHS, so a case mined below the commit window
-  whose truth file was touched again inside it is still contaminated — 48 of 50
-  cases on this repo, because neo's churn concentrates in exactly the files most
-  likely to be ground truth. `--drop-contaminated` empties the sample outright
-  here rather than cleaning it. Direction is the robust part; treat magnitudes as
-  bounded, and re-run on a CLEAN checkout of both trees, since a dirty working
-  tree puts every edited path into the same signal.
+  **Measure with `--no-git`, which is the default.** The scorer's recency signal
+  reads `git status --porcelain` plus the last 50 commits and holds PATHS, not
+  commits — so a case mined below the window whose truth file was touched again
+  inside it is still handed its own answer key, as is every file dirty in the tree
+  you are measuring from. `--skip-recent` does NOT fix this and a first version of
+  that docstring wrongly said it did. `neo --dry-run --no-git` gates `git_recent`
+  and nothing else (`_history_boost` and the rest of the re-rank stay live), which
+  is what `tools/rank_eval.py` had been doing all along. `--with-git` measures the
+  full pipeline and then reports `contaminated_cases` per run: with it on, 48 of 50
+  neo cases are contaminated, and the figures move by ≤0.03 — the leak is real but
+  was never what carried the result.
   **The re-rank is LOAD-BEARING, not redundant** (`pi_boost` + `_symbol_score` +
   `hist_boost`, applied after the first-pass score). Disabling it on the real CLI
   takes R@1 from 0.344 to **0.044** and MRR from 0.646 to **0.261**. An earlier
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -643,18 +643,29 @@
   this exact task, and BM25's `b` handles the concern with bounded, corpus-derived
   length normalization. Measured end-to-end over cases mined from git history
   (commit subject = query, changed non-test files = ground truth), R@10 / MRR:
-  neo 0.078→0.606 / 0.082→0.644, car 0.210→0.507 / 0.149→0.250, quip 0.213→0.857 /
-  0.188→0.698. `tools/rank_eval.py` is the harness.
-  **Two negative results, recorded so they are not retried.** RRF fusion with the
-  dense channel LOSES to BM25 alone (0.596 best-weighted vs 0.693) — dense returns
-  ~25 files against BM25's ~180 and is half as accurate — even after chunk allocation
-  was fixed. And a sweep of the re-rank weights (`pi_boost`, `_symbol_score`,
-  history: damped to 0.3, to 0.5, each disabled, content doubled) lands everywhere in
-  0.66–0.68, so the re-rank is *redundant* with content ranking rather than fighting
-  it. Four filename-tuning fixes were also measured and rejected; the filename is not
-  the evidence, the file is.
+  neo 0.078→0.603 / 0.082→0.655, car 0.210→0.487 / 0.149→0.233, quip 0.213→0.850 /
+  0.188→0.735, at `CONTENT_WEIGHT = 3.0`. `tools/rank_eval.py` is the harness, and
+  the numbers must be measured through the REAL CLI with `PYTHONPATH` pinned — see
+  the measurement warning above.
+  **The re-rank is LOAD-BEARING, not redundant** (`pi_boost` + `_symbol_score` +
+  `hist_boost`, applied after the first-pass score). Disabling it on the real CLI
+  takes R@1 from 0.344 to **0.044** and MRR from 0.646 to **0.261**. An earlier
+  version of this note claimed the opposite, from a weight sweep that landed
+  everywhere in 0.66–0.68 — measured at k=10, where every configuration is flat, and
+  through an in-process replica that omits the byte budget and adaptive limit, i.e.
+  exactly the stages that make the re-rank matter. Both errors are the ones
+  `tools/rank_eval.py` warns about in its own docstring. Per channel, `hist_boost`
+  contributes nothing measurable (identical results with it disabled) while
+  `_symbol_score` carries most of the effect.
+  RRF fusion with the dense channel LOSES to BM25 alone (0.596 best-weighted vs
+  0.693) — dense returns ~25 files against BM25's ~180 and is half as accurate. Treat
+  that as provisional: it was measured at k=10 before the same cutoff problem was
+  understood, and the docstring deferring it to a chunk-allocation fix is stale
+  because that fix landed in the same branch. Re-measure at k=3/k=5 before relying on
+  it. Four filename-tuning fixes were separately measured and rejected; the filename
+  is not the evidence, the file is.
   **A path named in the prompt is still pinned** (`EXPLICIT_PATH_BOOST=10.0`, chosen
-  to exceed every organic signal combined — content caps at +3.0, the re-rank boosts
+  to exceed every organic signal combined — content caps at +3.0 (`CONTENT_WEIGHT`), the re-rank boosts
   at +1.0 and +1.2). Without it a spelled-out path competed on generic filename-token
   overlap and lost: `src/neo/subcommands.py` ranked **163rd of 296** on a prompt
   naming it, below its own test file, because an 86KB file took a heavy size penalty. The file never
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -541,8 +541,9 @@
   (1) The venv installs neo EDITABLE against `src/`, so running
   `.venv/bin/python -m neo.cli` from a git worktree of another commit executes THIS
   tree's code against that tree's files — a "baseline" run that is not the baseline.
-  Every A/B needs `PYTHONPATH=<that tree>/src`. Measured: it made main look like
-  MRR 0.613 when its real figure is 0.082. (2) Calling `score_candidate` directly
+  Every A/B needs `PYTHONPATH=<that tree>/src`. Measured on the superseded harness
+  generation: it made main look like MRR 0.613 against a real figure of 0.082 (the
+  current harness puts main at 0.294 — different instrument, same trap). (2) Calling `score_candidate` directly
   measures the FIRST-PASS ranking only; `gather_context` then re-ranks with
   `pi_boost + hist_boost + _symbol_score` and applies an adaptive limit and a byte
   budget. A first-pass harness overstated R@10 by 0.14 against the real CLI. Validate
@@ -643,10 +644,23 @@
   this exact task, and BM25's `b` handles the concern with bounded, corpus-derived
   length normalization. Measured end-to-end over cases mined from git history
   (commit subject = query, changed non-test files = ground truth), R@10 / MRR:
-  neo 0.078→0.603 / 0.082→0.655, car 0.210→0.487 / 0.149→0.233, quip 0.213→0.850 /
-  0.188→0.735, at `CONTENT_WEIGHT = 3.0`. `tools/rank_eval.py` is the harness, and
-  the numbers must be measured through the REAL CLI with `PYTHONPATH` pinned — see
-  the measurement warning above.
+  neo 0.275→0.732 / 0.294→0.787, car 0.192→0.502 / 0.160→0.398, quip 0.175→0.720 /
+  0.166→0.658, at `CONTENT_WEIGHT = 3.0`. **`tools/rank_mine_eval.py` is the
+  harness** — not `tools/rank_eval.py`, which is a different instrument (12
+  hand-labelled prompts, this repo, recall@k, no MRR) and was named here in error
+  while the real one went uncommitted, leaving the figures unreproducible. Quoting
+  a number from one under the other's name is how `car` got into the record twice
+  at 0.969 and 0.507; keep the generation attached. An earlier generation of these
+  same figures read neo 0.078→0.603 / 0.082→0.655 — superseded, and not comparable,
+  because the harness that produced it no longer exists to re-run.
+  **The absolute values are upper bounds, and the tool says so per run.** The
+  scorer's recency signal holds PATHS, so a case mined below the commit window
+  whose truth file was touched again inside it is still contaminated — 48 of 50
+  cases on this repo, because neo's churn concentrates in exactly the files most
+  likely to be ground truth. `--drop-contaminated` empties the sample outright
+  here rather than cleaning it. Direction is the robust part; treat magnitudes as
+  bounded, and re-run on a CLEAN checkout of both trees, since a dirty working
+  tree puts every edited path into the same signal.
   **The re-rank is LOAD-BEARING, not redundant** (`pi_boost` + `_symbol_score` +
   `hist_boost`, applied after the first-pass score). Disabling it on the real CLI
   takes R@1 from 0.344 to **0.044** and MRR from 0.646 to **0.261**. An earlier
@@ -654,7 +668,8 @@
   everywhere in 0.66–0.68 — measured at k=10, where every configuration is flat, and
   through an in-process replica that omits the byte budget and adaptive limit, i.e.
   exactly the stages that make the re-rank matter. Both errors are the ones
-  `tools/rank_eval.py` warns about in its own docstring. Per channel, `hist_boost`
+  `tools/rank_mine_eval.py` warns about in its own docstring (`rank_eval.py` was
+  cited here and carries neither warning). Per channel, `hist_boost`
   contributes nothing measurable (identical results with it disabled) while
   `_symbol_score` carries most of the effect.
   RRF fusion with the dense channel LOSES to BM25 alone (0.596 best-weighted vs
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -536,6 +536,17 @@
   asked git for the same remote 3× — ~100 forks/cycle, ~26k/week. Staleness is
   bounded to one cycle because a stale hit would write facts under the wrong
   `project_id`.
+- **Measuring retrieval changes: pin `PYTHONPATH`, and measure the REAL pipeline.**
+  Two traps, both hit during the BM25 work, both producing confident wrong numbers.
+  (1) The venv installs neo EDITABLE against `src/`, so running
+  `.venv/bin/python -m neo.cli` from a git worktree of another commit executes THIS
+  tree's code against that tree's files — a "baseline" run that is not the baseline.
+  Every A/B needs `PYTHONPATH=<that tree>/src`. Measured: it made main look like
+  MRR 0.613 when its real figure is 0.082. (2) Calling `score_candidate` directly
+  measures the FIRST-PASS ranking only; `gather_context` then re-ranks with
+  `pi_boost + hist_boost + _symbol_score` and applies an adaptive limit and a byte
+  budget. A first-pass harness overstated R@10 by 0.14 against the real CLI. Validate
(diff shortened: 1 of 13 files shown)


JLiotta-25:
## Correction after review: the shipped weight was not the documented weight

Linus found a blocker I shipped **and measured** without noticing, plus a wrong conclusion I had already written into CLAUDE.md. Both fixed in 7d00723. Every number in the description above was produced at a weight nobody chose.

### The content term was applied twice

`score += 3.0 * content_relevance` appeared at two adjacent lines, so the effective weight was **6.0** while every docstring, comment, commit message and table said 3.0. Introduced by the cherry-pick that merged #188's scorer with this one — it kept both comment blocks, and they contradict each other (one says the tie-breakers sum to 1.3, the other 1.75).

Now a named `CONTENT_WEIGHT` applied once. Re-measured end-to-end at both weights on all three repos: **W=3.0 wins** (mean MRR 0.541 vs 0.531, mean R@3 0.471 vs 0.464). The documented value turned out to be the better one, which is luck, not process.

**Corrected results** (W=3.0, real CLI, `PYTHONPATH` pinned per tree):

| repo | R@10 | MRR |
|---|---|---|
| neo | 0.078 → **0.603** | 0.082 → **0.655** |
| car | 0.210 → **0.487** | 0.149 → **0.233** |
| quip | 0.213 → **0.850** | 0.188 → **0.735** |

### "The re-rank is redundant" was wrong

Disabling `pi_boost` + `_symbol_score` + `hist_boost` on the real CLI takes **R@1 from 0.344 to 0.044** and MRR from 0.646 to 0.261. It is load-bearing.

I concluded the opposite from a weight sweep measured **at k=10** — where every configuration on this repo is flat — run through an in-process replica that omits the byte budget and adaptive limit, i.e. exactly the stages that make the re-rank matter. Both mistakes are the ones `tools/rank_eval.py` warns about *in its own docstring, in this branch*. Removed from CLAUDE.md. The surviving half is real and kept: `hist_boost` contributes nothing measurable.

The fusion negative is downgraded to **provisional** for the same reason (also k=10), and its stated mechanism was stale — it deferred to a chunk-allocation fix that landed two commits later in this same branch.

### Four constants survived a mutation sweep

One of them was the shipped bug. The test that should have caught it named 3.0 in its docstring and asserted an inequality satisfied by anything above 1.76 — naming a number in prose while asserting a direction is how a constant goes unpinned and looks pinned. All four are now pinned by value, plus a structural test that the term is applied exactly once. All six mutants re-run and caught.

### Binary files were entering the BM25 corpus

`errors="ignore"` turned PDFs into junk tokens: **5 files were 1.7% of documents and 32.9% of all tokens**, and they set `avgdl` — so BM25's length normalization depended on how many papers happened to sit in the repo. A size term driven by something unrelated to relevance, re-entering through the corpus statistics. NUL-byte probe added; `avgdl` 2773 → 1824.

### Smaller

- `MAX_INDEXED_BYTES` → `MAX_INDEXED_CHARS`. `TextIOWrapper.read(n)` counts code points, and CLAUDE.md already records this exact lie in `_render_context_files`.
- The chunk-cap console line no longer promises an unconditional floor that `_allocate_slots` does not always give.
- Three conflicting measurement tables reconciled to one, with the harness generation named — `car` had appeared as both 0.969 and 0.507 for the same quantity.
- BM25's O(N) memory and query profile deferred and filed as #195 (~15 s and ~500 MB at 5,000 files; the fix belongs in `memory/bm25.py`, which has other callers).

kkeeling:
## Rebased onto `main` (`5bbee46`), re-measured, evidence below

**Rebase.** 16 commits replayed onto `origin/main` with **zero conflicts**, Jennifer's authorship preserved on every one (`git log --format='%an'` → `Jennifer Liotta <jliotta@r337.com>` ×16). GitHub reported `CONFLICTING` against a *merge*; the *rebase* is clean. The apparent add/add on `tools/rank_eval.py` resolves itself: this branch's first commit created that file, #192 shipped exactly that version to `main`, and git recognised the change as already upstream.

One commit added on top (mine, not Jennifer's): `45fa895 docs(baselines): name the instrument by revision, not by path` — see "Instrument note" at the bottom.

---

### M1 — before/after, all three flagships

Same instrument on both arms: `5bbee46:tools/rank_eval.py`, the #192 harness the baselines doc canonicalises. Same mined case files on both arms (mined once per repo; mining reads git history only). **Before** reproduces `docs/eval-baselines-2026-08.md` digit-for-digit — 214/225/173 cases mined, 209/221/173 evaluable, every published figure identical — so the two arms differ only in the neo tree under test.

`neo_current` is the shipped selector. `bm25_content` is the doc's reference lane and is *unchanged by this PR* (it re-ranks independently of `score_candidate`); it is kept in the table as a fixed yardstick.

**neo** — 209 evaluable cases

| strategy | R@1 | R@3 | R@5 | R@10 | R@20 | MRR |
|---|---|---|---|---|---|---|
| `neo_current` **before** | 0.043 | 0.098 | 0.146 | 0.212 | 0.418 | 0.136 |
| `neo_current` **after** | **0.455** | **0.630** | **0.681** | **0.751** | **0.878** | **0.714** |
| `bm25_content` (yardstick) | 0.266 | 0.541 | 0.612 | 0.695 | 0.740 | 0.559 |

**aieweb** — 221 evaluable cases

| strategy | R@1 | R@3 | R@5 | R@10 | R@20 | MRR |
|---|---|---|---|---|---|---|
| `neo_current` **before** | 0.053 | 0.117 | 0.182 | 0.244 | 0.296 | 0.180 |
| `neo_current` **after** | **0.470** | **0.653** | **0.707** | **0.774** | **0.839** | **0.759** |
| `bm25_content` (yardstick) | 0.417 | 0.683 | 0.757 | 0.823 | 0.880 | 0.726 |

**m365dotnet** — 173 evaluable cases *(the gate: plan target is MRR ≥ 0.60)*

| strategy | R@1 | R@3 | R@5 | R@10 | R@20 | MRR |
|---|---|---|---|---|---|---|
| `neo_current` **before** | 0.006 | 0.029 | 0.035 | 0.097 | 0.146 | 0.051 |
| `neo_current` **after** | **0.419** | **0.673** | **0.767** | **0.883** | **0.929** | **0.738** |
| `bm25_content` (yardstick) | 0.284 | 0.559 | 0.684 | 0.813 | 0.877 | 0.607 |

**Headline**

| Repo | MRR before → after | recall@10 before → after |
|---|---|---|
| neo | 0.136 → **0.714** (5.3×) | 0.212 → **0.751** |
| aieweb | 0.180 → **0.759** (4.2×) | 0.244 → **0.774** |
| **m365dotnet** | 0.051 → **0.738** (14.5×) | 0.097 → **0.883** |

m365dotnet clears the ≥ 0.60 gate at **0.738**, and clears it by beating plain content-BM25 (0.607) rather than merely matching it — the test-file demotion and the retained filename/entry-point signals add on top of the content lane rather than cancelling it. No flagship regresses on any cutoff.

`dense` / `rrf_bm25_dense` stay unbaselined on both arms for the reason the doc gives (`ProjectIndex.chunks` empty on all three repos); the harness prints 0.000 for `dense`, which is not a measurement.

---

### G-invariant battery — post-#194, m365dotnet

Canonical six-prompt battery via `tools/m2_battery.sh`, then the doc's G1 script over the per-prompt selections.

| prompt | distinct files | `git check-ignore` excludes | duplicate copies | inside `.worktrees/` | unresolvable |
|---|---|---|---|---|---|
| P1 | 22 | 0 | 0 | 0 | 0 |
| P2 | 23 | 0 | 0 | 0 | 0 |
| P3 | 24 | 0 | 0 | 0 | 0 |
| P4 | 22 | 0 | 0 | 0 | 0 |
| P5 | 24 | 0 | 0 | 0 | 0 |
| P6 | 20 | 0 | 0 | 0 | 0 |
| **BATTERY UNION** | **111** | **0** | **0** | **0** | **0** |

- **G1-inv holds** — zero gitignored selections, zero duplicate copies, on the repo carrying 44 in-repo worktrees.
- **Named-file guarantee unchanged** — P1's `src/Parslee.M365.Api/Program.cs` and P2's `src/Parslee.M365.Api/Controllers/EntitlementsController.cs` are both present in their prompt's selection.
- Coverage moved from a 141-file union to a **111**-file union at the same 180 context entries. That is concentration, not loss: the per-file character cap means the selector now spends more of a fixed budget on fewer, better-ranked files, which is the same effect the M1 R@1 jump measures.

---

### M2 — reported, and it regresses

| | baseline (`77251df`) | post-#194 | target |
|---|---|---|---|
| battery median wall | 10.54 s | **36.41 s** (3.5×) | ≤ 5 s |
| peak `ru_maxrss` | 1368.0 MiB | **1368.4 MiB** (flat) | ≤ 500 MB |

Stated plainly rather than buried: reading every candidate's content on every call costs **26 s per invocation** on m365dotnet. Memory is unchanged. This is the cost the unified-store plan already priced in — the plan's pre-measurement projection was ~27 s post-#194, and **Goal 6** ("move #194's per-call BM25 into the on-disk store beside the semantic catalog") exists to pay it back. M2 is Goal 6's DoD, not this PR's; Goal 2's gate is M1 plus the G-invariant battery, both above. Flagging it here so the number is on the record before Goal 6 quotes it as its own baseline.

---

### Tests

`pytest` on the rebased branch: **2246 passed, 8 skipped, 1 failed** in 601.76 s.

The one failure is `tests/test_integration.py::TestPerformanceIntegration::test_consolidation_performance_with_all_boosts` — a wall-clock assertion (`elapsed_ms < 100`, observed 173 ms). It is **not a regression from this PR**: the same test on unmodified `main`, same machine, same venv shape, fails at **202 ms**. It is the exact defect class #183/#185 already fixed for the learning-loop benchmark — a latency threshold used as a correctness gate — surviving in a different file. Out of scope here; CI is green on `main`, so it does not reproduce on the runner.

---

### Instrument note (the one added commit)

This PR gives `tools/rank_eval.py` back to the 12-prompt hand-labelled recall harness and moves the git-mined cross-repo MRR harness to `tools/rank_mine_eval.py`. `docs/eval-baselines-2026-08.md`'s "Reproduce M1" block names `tools/rank_eval.py` by path, so on merge that block would point at a file that no longer accepts `--build-from-git`/`--eval` and no longer reports MRR. `45fa895` pins the instrument by revision (`5bbee46:tools/rank_eval.py`) and records the two arguments `score_candidate` gained, which is all a later goal needs to re-run the published tables against a post-#194 tree. It deliberately does **not** redirect the block to `rank_mine_eval.py` — different instrument, different absolute numbers, and that substitution is the failure `rank_mine_eval.py`'s own header warns about.

### Exact commands

```bash
platform_root=<platform-root>
git show 5bbee46:tools/rank_eval.py > /tmp/canonical_rank_eval.py

# cases (identical for both arms)
for r in neo aieweb m365dotnet; do
  .venv-eval/bin/python /tmp/canonical_rank_eval.py --build-from-git $platform_root/$r > /tmp/cases_$r.json
done

# before: run from a main worktree's venv; after: from this branch's venv,
# with the score_candidate call updated as described in 45fa895
for r in neo aieweb m365dotnet; do
  NEO_OBSERVER_AUTOSTART=0 .venv-eval/bin/python /tmp/canonical_rank_eval.py \
    --eval $platform_root/$r /tmp/cases_$r.json
done

# battery + G1
tools/m2_battery.sh "$PWD/.venv-eval/bin/neo" $platform_root/m365dotnet /tmp/m2_m365_after
```

Repos measured at the baselines doc's recorded HEADs, verified unchanged: neo `5bbee46` (src identical to the doc's `77251df`), aieweb `26fff07`, m365dotnet `61dc4a1`. Machine: Apple M4 Max, 64 GB, macOS 26.5.2, Python 3.13.7 — same as the baselines doc.

<sub>Goal 2 of `docs/unified-store-plan.md` — Land the queue.</sub>

kkeeling (approved):
Rebased onto `main` (`5bbee46`) with zero conflicts and Jennifer's authorship preserved on all 16 commits. CI green on 3.10-3.13.

M1 re-measured on the same instrument for both arms (`5bbee46:tools/rank_eval.py`), with the before arm reproducing `docs/eval-baselines-2026-08.md` digit-for-digit:

| Repo | MRR before → after | recall@10 before → after |
|---|---|---|
| neo | 0.136 → 0.714 | 0.212 → 0.751 |
| aieweb | 0.180 → 0.759 | 0.244 → 0.774 |
| m365dotnet | 0.051 → **0.738** | 0.097 → **0.883** |

m365dotnet clears the plan's ≥ 0.60 gate at 0.738, above plain content-BM25 (0.607). G-invariant battery on m365dotnet: 0 gitignored, 0 duplicates, 0 unresolvable across the 111-file union; both prompt-named paths pinned. M2 regresses 10.54 s → 36.41 s, which is the cost Goal 6 exists to repay — full numbers in the evidence comment above.

Approving under the merge authority recorded in `docs/unified-store-plan.md` (Keenan, 2026-08-11): green CI + pasted evidence.

PR #201, opened 2026-08-11
docs: unified-store Goal 1 — flagship baselines

kkeeling:
## Description

Publishes the M1 / M2 / G1-invariant baselines for the [unified-store plan](https://github.com/Parslee-ai/neo/pull/200) — "Goal 1. Trailhead" — measured on the three flagship repos with **neo v0.44.0 at `main` = `77251df`**.

Adds two files, both docs/tools:

- `docs/eval-baselines-2026-08.md` — every number, the exact command that produced it, the repo SHAs measured, machine caveats, and an explicit list of what could **not** be measured and why.
- `tools/m2_battery.sh` — the canonical M2 prompt battery, committed as an artifact so later goals re-run the same six prompts instead of re-keying them from prose.

### Headline numbers

| Metric | neo | aieweb | m365dotnet | Target |
|---|---|---|---|---|
| **M1** MRR (`neo_current`) | 0.136 | 0.180 | **0.051** | m365dotnet ≥ 0.60 |
| **M1** recall@10 (`neo_current`) | 0.212 | 0.244 | **0.097** | — |
| **M2** wall-clock, median | — | — | **10.54 s** | ≤ 5 s |
| **M2** peak RSS (`ru_maxrss`) | — | — | **1.43 GB** | ≤ 500 MB |
| **G1-inv** gitignored files selected | — | — | **0 / 141** | 0 |
| **G1-inv** duplicate copies selected | — | — | **0 / 141** | 0 |

Same-process comparison on the identical m365dotnet eval set, which is the number that matters most:

| strategy | R@1 | R@5 | R@10 | MRR |
|---|---|---|---|---|
| `neo_current` | 0.006 | 0.035 | 0.097 | **0.051** |
| `bm25_content` | 0.284 | 0.684 | 0.813 | **0.607** |

Three readings:

1. **Retrieval on the largest flagship is the weakest.** m365dotnet lands a correct file in the top 10 for about one query in ten; MRR is 11.8× below target. Not a tuning gap.
2. **The gap is not "retrieval is hard here."** Plain BM25 over content scores 0.607 MRR on the same cases, in the same process, on the same machine. The signal is in the repo; the shipped scorer isn't using it.
3. **G1-inv holds post-#186.** The 2026-08-10 reality was 14/16 ignored and 12/16 duplicates. Measured across the whole battery on the repo carrying 44 in-repo worktrees: **0 and 0**, over all 141 distinct files selected.

Also measured: M2 cost is flat across prompt shape (9.69 s → 11.26 s from cheapest to dearest, RSS varying under 2 MiB). Naming a file costs the same as naming nothing — the signature of a fixed scan-the-world cost, which is the premise the whole plan rests on, now measured rather than asserted.

## Type of Change

- [x] Documentation update

## Related Issue

Unified Store Plan Goal 1 (Trailhead) — plan of record in #200. No issue number to close.

## Motivation and Context

Every later goal in the plan has to show its numbers against a baseline, and until now m365dotnet and aieweb were unmeasured. This is that baseline. The mining parameters and the six-prompt M2 battery are declared **canonical** in the doc: a later goal that changes any of them must re-baseline all three repos, not just the one it cares about.

## How Has This Been Tested?

This is a measurement PR — the "test" is that every number is reproducible from a command printed next to it.

- [x] M1: `tools/rank_eval.py --build-from-git` then `--eval` on all three repos (209 / 221 / 173 evaluable cases)
- [x] M2: `tools/m2_battery.sh`, 1 discarded warm-up + 3 timed runs × 6 prompts = 18 timed runs
- [x] G1-inv: `git check-ignore --stdin` + SHA-256 content grouping over all **141** distinct selected files (the battery union)
- [x] `ruff check src/ tests/ tools/` — all checks passed
- [x] `shellcheck tools/m2_battery.sh` — clean

**Test Configuration**:
- Python version: 3.13.7
- OS: macOS 26.5.2 (Apple M4 Max, 16 cores, 64 GB)
- Neo version: 0.44.0 at `77251dff54d0726fb7d98e6e6daa525e1b8b2c6d`, installed editable into a dedicated venv (**not** the globally installed `neo`, which is a different version on this machine)

## Checklist

- [ ] I have added tests that prove my fix is effective or that my feature works — n/a, docs + a measurement script; no runtime code changed
- [x] New and existing unit tests pass locally with my changes — no Python source touched; lint green

## Additional Notes

**What is deliberately NOT baselined**, stated in the doc rather than estimated (the goal's standing constraint is: if a number can't be produced, say so with the blocking reason, never interpolate):

- **`dense` / `rrf_bm25_dense` M1 rows.** The harness prints `0.000`, and that is *not* a measurement — `ProjectIndex(root).chunks` was empty for all three repos, and `rrf` degrades to the BM25 ranking when the dense list is empty. Building the index is blocked two ways: it persists to `<repo_root>/.neo/`, i.e. *inside* the measured repo, which this goal may not write to for aieweb/m365dotnet; and the neo-only build was killed at **11 min 21 s** holding **~9.97 GB RSS** with no output. That second observation lands squarely on Goal 6 and Goal 7 — a cold semantic-index build of the *smallest* flagship didn't finish in 11 minutes.
- **`origin/main` SHAs for aieweb and m365dotnet.** Both hold uncommitted in-flight work (39 and 1 modified files), and checking them out is a write this goal is forbidden to make. They were measured in place at their exact HEADs, recorded in the doc so a later goal can re-measure the identical tree. m365dotnet is 166 commits behind `origin/main`, aieweb 85.

**Measurement footgun recorded in the doc**, because it produced the right answer from a broken measurement: `--dry-run` emits windowed files as `<path> (lines a-b) - N bytes`, and a first extraction pass left the `(lines …)` suffix on 23 of 180 paths. `git check-ignore` then silently tested paths that don't exist and still returned `0 ignored`. The battery script strips the suffix, and the doc's G1 table carries an `unresolvable paths` column so that class of error announces itself next time.

**Verification that nothing leaked outside neo:** m365dotnet and aieweb are byte-identical to their pre-measurement state (same HEAD, same dirty counts, no `.neo/` artifacts). No runtime or product code changed — `docs/` and `tools/` only.

**Merge authority:** #200 has since merged, so the plan doc's ruling is now on `main`: *"goals open PRs; auto-merge on green gates is authorized (Keenan, 2026-08-11)"* (`docs/unified-store-plan.md:18`). Per this goal's constraint that makes merge authorized on green CI with the numbers pasted, both of which now hold. This agent does not merge its own PR — leaving the merge to the maintainer.

---

## Update — fresh non-author verification pass

A fresh non-author Claude independently re-ran the M1 evals for **neo and aieweb** (both reproduce to the last digit), recomputed every M2 aggregate from the raw `/usr/bin/time -l` captures (all correct), re-verified the G1 claim two ways, and confirmed the `dense`/RRF "not measured" characterization against the harness source. It found **no wrong measurement** — but four reporting defects, now fixed in `85754a9`:

1. **The G1 denominator was inflated.** The headline said `0 / 171`, but 171 was the *sum* of the six per-prompt distinct counts, which counts a file once per prompt that selected it. The battery's true distinct **union is 141** — 30 files were selected by more than one prompt. Corrected throughout. **The invariant verdict does not change**: 0 ignored / 0 duplicates, verified over both the 141 union and the 171 with-repeats set.
2. **"180 entries cover 171 distinct files" was false** (141). ~~and its "nine slots are a second window" was a per-prompt number presented as battery-wide (23 battery-wide)~~ — **that half of this correction was itself wrong and is reverted in `d64d47a`; see the round-2 note below.**
3. **The three M1 tables printed numeric `rrf_bm25_dense` values** while the prose below called both dense rows "not measured". Those values are the BM25 row copied by the `dn == []` fallback, not an RRF result. Now marked *not measured*, matching the prose.
4. **"2.7× over on memory" divided MiB by an MB-stated target.** Correct ratio is **2.87×**; the base is named at the point of use now.

Also closed a hole in the reproduction contract the verifier flagged: the M2 aggregates were computed by hand and lived in no committed command. `tools/m2_battery.sh` now derives them itself (median wall, peak RSS, target ratios, and all three file counts) and writes `union.files` for the G1 check.

Corrected headline G1 row: **0 gitignored / 0 duplicate copies, out of 141 distinct files selected across the battery.**

---

## Update — round 2, adversarial review resolved

`neo` is **not** in the `car-pr-review` daemon's coverage list, so no Grace verdict will ever land here. Per the out-of-coverage path, a fresh non-author Claude reviewed the PR and posted its findings: [Adversarial Review (local fresh-context reviewer)](https://github.com/Parslee-ai/neo/pull/201#issuecomment-5261157625) — **VERDICT: CHANGES REQUESTED, 1 blocking.** It was right. Fixed in `d64d47a`.

**The blocking finding.** The G1 prose said "23 of the 180 entries are a second window of a file already selected." 23 is the count of *windowed* entries — the number the footgun paragraph cites correctly. The repeat-entry counts are different quantities, recomputed here from the raw artifacts:

| quantity | value | derivation |
|---|---|---|
| context entries | 180 | 6 prompts × 30 |
| sum of per-prompt distinct | 171 | 29+28+29+29+28+28 |
| **battery union distinct** | **141** | the number to quote |
| repeat entries, battery-wide | **39** | 180 − 141 |
| repeat entries, within-prompt | **9** | 180 − 171 |
| files picked by >1 prompt | **21** | (accounting for 30 excess selections) |

The old sentence also contradicted the "overstates battery coverage by 28%" claim three lines below it, which is computed from 39. And the round-1 Update above made it worse — it presented 23 as a *correction* of 9, when 9 was the right within-prompt figure all along. Both are now stated with their derivations, and the script prints all three so the next person does not have to trust prose.

Second fix: "30 files were picked by more than one prompt" was 30 excess *selections* spread over **21 files**.

Neither touches a measurement. **G1-inv is unchanged: 0 gitignored / 0 duplicate copies over the 141-file union.**

**Beyond the findings — the reproduction contract is now actually proven.** The published tables came from an ad-hoc first pass; `tools/m2_battery.sh` was written afterwards. Offering a script as the reproduction contract without ever running the committed version is a gap, so it was run end-to-end into a clean directory (3 min 28 s):

| quantity | first pass | committed-script re-run |
|---|---|---|
| entries / per-prompt distinct / **union** | 180 / 171 / **141** | 180 / 171 / **141** |
| the union file set itself | — | **byte-identical** |
| repeat entries battery-wide / within-prompt | 39 / 9 | 39 / 9 |
| G1-inv (ignored / dup / `.worktrees/` / unresolvable) | 0 / 0 / 0 / 0 | 0 / 0 / 0 / 0 |
| peak `ru_maxrss` | 1368.0 MiB (2.87×) | 1368.0 MiB (2.87×) |
| battery median wall | **10.54 s** | 10.22 s |

Selection is deterministic; wall-clock is not. The entire 3% wall-clock move is one scheduler stall (P4 run 1 at 19.27 s against 10.18 s on run 3 of the same prompt) — which is why the protocol reports a median. **The published baseline stays 10.54 s**; re-baselining to the second sample would be picking a number for no reason. The doc now states the ±0.5 s noise band so a later goal does not read noise as a result.

Also closed from the non-blocking list: `results.csv` emits `entries` **and** `distinct_files` rather than one ambiguous `files_selected` column (which read 29 where the doc's table said 30); the measurement loop now aborts loudly instead of writing an empty CSV cell the aggregation would read as a number (and the pipeline status is checked *after* the `tee`, since the loop runs in a subshell and its `exit` would otherwise be swallowed); and `<platform-root>` / `<neo-worktree>` are defined instead of left as undefined placeholders in pasteable commands.

`shellcheck tools/m2_battery.sh` clean · `ruff check src/ tests/ tools/` all checks passed.

### DIFF
--- a/docs/eval-baselines-2026-08.md
+++ b/docs/eval-baselines-2026-08.md
@@ -12,6 +12,17 @@ Every number here was produced by running the command printed beside it. Where a
 number could not be produced, the row says so and gives the blocking reason —
 nothing in this document is estimated, projected, or interpolated.
 
+**Two placeholders appear in the commands below**, deliberately, so the doc is not
+pinned to one developer's home directory. `<platform-root>` is the `parslee-knowledge`
+checkout that holds the child repos as subdirectories (`<platform-root>/m365dotnet`,
+`<platform-root>/aieweb`, `<platform-root>/neo`); `<neo-worktree>` is the neo worktree
+this document was produced from. Set them once before pasting anything:
+
+```bash
+export platform_root=/path/to/parslee-knowledge     # substitute for <platform-root>
+export neo_worktree=$platform_root/neo/.worktrees/g_msp4yrhu_d974cb-unified-store-goal-1
+```
+
 ---
 
 ## Headline
@@ -302,6 +313,34 @@ It also writes into the output directory: each run's raw `/usr/bin/time -l` capt
 `results.csv`, and `union.files` — the distinct union across the battery, which is the
 input to the G1 check below.
 
+### Reproducibility re-run — 2026-08-12
+
+The tables above came from an ad-hoc first pass; `tools/m2_battery.sh` was written
+afterwards to make that pass a command. A doc that publishes a script as its
+reproduction contract owes you evidence that the *committed script* produces the
+published numbers, so it was run end-to-end a second time into a clean directory
+(3 min 28 s wall). What it showed, split by what is deterministic and what is not:
+
+| quantity | first pass | committed-script re-run | verdict |
+|---|---|---|---|
+| context entries / per-prompt distinct / **union** | 180 / 171 / **141** | 180 / 171 / **141** | identical |
+| per-prompt distinct (P1…P6) | 29, 28, 29, 29, 28, 28 | 29, 28, 29, 29, 28, 28 | identical |
+| the union file set itself | — | byte-identical set | identical |
+| repeat entries: battery-wide / within-prompt | 39 / 9 | 39 / 9 | identical |
+| files picked by >1 prompt | 21 | 21 | identical |
+| G1-inv: ignored / duplicate / `.worktrees/` / unresolvable | 0 / 0 / 0 / 0 | 0 / 0 / 0 / 0 | identical |
+| peak `ru_maxrss` | 1368.0 MiB (2.87×) | 1368.0 MiB (2.87×) | identical |
+| **battery median wall** | **10.54 s** | **10.22 s** | ±3%, see below |
+
+**Selection is deterministic; wall-clock is not.** Every file-count, the G1 verdict and
+the RSS peak land on the same value twice. Wall-clock moved 3%, and the whole of that
+move is one scheduler stall: P4 run 1 took 19.27 s in the re-run against a 10.18 s
+run 3 on the same prompt. That is precisely why the protocol reports a median and why
+this document says to treat M2 as a same-machine before/after instrument. **The
+published baseline stays 10.54 s** — re-baselining to the second sample would be
+picking a number for no reason. A later goal comparing against it should read a
+change under ~±0.5 s as noise, not as a result.
+
 ---
 
 ## G1-invariant on m365dotnet
@@ -331,7 +370,8 @@ worst duplicate exposure on this machine — and it holds computed either way, o
 
 **Read the union row, not the sum row.** 171 is the sum of the six per-prompt distinct
 counts, which counts a file once per prompt that selected it; the battery selected
-**141** distinct files in total, so 30 files were picked by more than one prompt. The
+**141** distinct files in total. The 30-file difference is 30 *excess selections*
+spread over **21 files** that more than one prompt picked — not 30 files. The
 sum row is kept only to make the double-count visible. A later goal that reports a true
 union and compares it to 171 will see a 17% "coverage regression" that never happened.
 
@@ -351,13 +391,18 @@ Definitions, so a later re-measure counts the same things:
 
 Two observations that are *not* G1 violations but belong on the record:
 
-- **180 context entries cover 141 distinct files.** Battery-wide, 23 of the 180 entries
-  are a second window of a file already selected — `MAX_CHUNKS_PER_FILE` working as
-  designed, not duplication. Within a single prompt the effect is smaller (30 entries →
-  28–29 distinct). So "30 files selected" overstates that prompt's coverage by ~5%, and
-  a raw entry count overstates battery coverage by 28%. Any later goal reporting a file
-  count must say which of the three it is counting: entries, per-prompt distinct, or
-  union.
+- **180 context entries cover 141 distinct files.** Battery-wide, **39** of the 180
+  entries repeat a file some prompt had already selected (180 − 141) — a mix of
+  `MAX_CHUNKS_PER_FILE` second windows and the same file being picked by more than one
+  prompt. Neither is duplication in the G1 sense. Within a single prompt the effect is
+  far smaller: only **9** entries battery-wide are a second window of a file already
+  selected *by that same prompt* (180 − 171), which is what turns 30 entries into 28–29
+  distinct. So "30 files selected" overstates that prompt's coverage by ~5%, and a raw
+  entry count overstates battery coverage by 28%. Any later goal reporting a file count
+  must say which of the three it is counting: entries, per-prompt distinct, or union.
+  (Do not confuse either number with the **23** *windowed* entries in the footgun note
+  below — that is the count of entries carrying a `(lines a-b)` range, a different
+  quantity that happens to sit in the same range.)
 - **Named paths were pinned.** P1's `Program.cs` and P2's `EntitlementsController.cs`
   each appear (2 windows each). That is a G2-inv spot check, not a G2-inv
   measurement — the full G2 battery is Goal 3's deliverable.
--- a/docs/eval-baselines-2026-08.md
+++ b/docs/eval-baselines-2026-08.md
@@ -22,8 +22,12 @@ nothing in this document is estimated, projected, or interpolated.
 | **M1** recall@10 (`neo_current`) | 0.212 | 0.244 | **0.097** | — |
 | **M2** wall-clock, median | — | — | **10.54 s** | ≤ 5 s |
 | **M2** peak RSS (`ru_maxrss`) | — | — | **1.43 GB** | ≤ 500 MB |
-| **G1-inv** gitignored files selected | — | — | **0 / 171** | 0 |
-| **G1-inv** duplicate copies selected | — | — | **0 / 171** | 0 |
+| **G1-inv** gitignored files selected | — | — | **0 / 141** | 0 |
+| **G1-inv** duplicate copies selected | — | — | **0 / 141** | 0 |
+
+141 is the **union** of distinct files the selector put in context across all six
+battery prompts (180 context entries in total). Count the union, never the sum of
+per-prompt counts — see the G1 section.
 
 Three readings, stated plainly:
 
@@ -136,7 +140,7 @@ three repos, not just the one it cares about.
 | `neo_current` | 0.043 | 0.098 | 0.146 | 0.212 | 0.418 | 0.136 |
 | `bm25_content` | 0.266 | 0.541 | 0.612 | 0.695 | 0.740 | 0.559 |
 | `dense` | *not measured* | | | | | |
-| `rrf_bm25_dense` | 0.266 | 0.541 | 0.612 | 0.695 | 0.740 | 0.559 |
+| `rrf_bm25_dense` | *not measured* | | | | | |
 
 **aieweb** — 221 evaluable cases
 
@@ -145,7 +149,7 @@ three repos, not just the one it cares about.
 | `neo_current` | 0.053 | 0.117 | 0.182 | 0.244 | 0.296 | 0.180 |
 | `bm25_content` | 0.417 | 0.683 | 0.757 | 0.823 | 0.880 | 0.726 |
 | `dense` | *not measured* | | | | | |
-| `rrf_bm25_dense` | 0.417 | 0.683 | 0.757 | 0.823 | 0.880 | 0.726 |
+| `rrf_bm25_dense` | *not measured* | | | | | |
 
 **m365dotnet** — 173 evaluable cases
 
@@ -154,7 +158,7 @@ three repos, not just the one it cares about.
 | `neo_current` | 0.006 | 0.029 | 0.035 | 0.097 | 0.146 | 0.051 |
 | `bm25_content` | 0.284 | 0.559 | 0.684 | 0.813 | 0.877 | 0.607 |
 | `dense` | *not measured* | | | | | |
-| `rrf_bm25_dense` | 0.284 | 0.559 | 0.684 | 0.813 | 0.877 | 0.607 |
+| `rrf_bm25_dense` | *not measured* | | | | | |
 
 Harness cost, for scheduling later runs: neo 22.8 s / 131 MB, aieweb 45.1 s / 165 MB,
 m365dotnet 275.5 s / 800 MB.
@@ -174,7 +178,8 @@ for r in ['neo','aieweb','m365dotnet']:
 # neo 0 / aieweb 0 / m365dotnet 0
 ```
 
-Because `rrf_bm25_dense` degrades to `bmr` when `dn` is empty, its row is a duplicate
+Because `rrf_bm25_dense` degrades to `bmr` when `dn` is empty, its row would be a
+byte-for-byte duplicate
 of `bm25_content` and carries no independent information either. Both rows are
 recorded as **not measured**, not as zero.
 
@@ -264,8 +269,10 @@ background observer sweep cannot land inside a timed run.
 **Battery: median wall-clock 10.54 s** (n = 18, min 7.45, max 11.97).
 **Battery: peak `ru_maxrss` 1,434,402,816 bytes = 1368.0 MiB = 1.43 GB.**
 
-Against the M2 targets of ≤ 5 s and ≤ 500 MB: **2.1× over on time, 2.7× over on
-memory.**
+Against the M2 targets of ≤ 5 s and ≤ 500 MB: **2.1× over on time, 2.87× over on
+memory.** (1,434,402,816 B ÷ 500 MB. The plan states the target in MB, so the ratio is
+computed in MB — reading `ru_maxrss` as MiB against a MiB target gives 2.74× and
+understates the overage by 5%. Pick one base; this document uses the plan's.)
 
 Two things worth noting about the shape of the result:
 
@@ -285,8 +292,15 @@ cd <neo-worktree>
 tools/m2_battery.sh "$PWD/.venv-eval/bin/neo" <platform-root>/m365dotnet /tmp/m2_m365
 ```
 
-Prints one CSV row per timed run and writes each run's raw `/usr/bin/time -l` output
-and selected-file list into the output directory.
+Prints one CSV row per timed run, then prints the **aggregates published above** —
+per-prompt median/min/max wall, per-prompt peak RSS, the battery median and peak, and
+the ratios against the M2 targets. The aggregation is in the script rather than done by
+hand, so every number in the M2 tables comes out of this one command.
+
+It also writes into the output directory: each run's raw `/usr/bin/time -l` capture
+(`<id>_run<n>.txt`), each run's selected-file list (`<id>_run<n>.files`), the raw
+`results.csv`, and `union.files` — the distinct union across the battery, which is the
+input to the G1 check below.
 
 ---
 
@@ -308,10 +322,18 @@ worktrees** under `.worktrees/`, which is exactly the condition that produced th
 | P4 | 30 | 29 | 0 | 0 | 0 | 0 |
 | P5 | 30 | 28 | 0 | 0 | 0 | 0 |
 | P6 | 30 | 28 | 0 | 0 | 0 | 0 |
-| **TOTAL** | **180** | **171** | **0** | **0** | **0** | **0** |
+| *sum of rows* | 180 | 171 | 0 | 0 | 0 | 0 |
+| **BATTERY UNION** | **180** | **141** | **0** | **0** | **0** | **0** |
 
 **G1-inv holds.** Post-#186 this is not "near zero", it is zero, on the repo with the
-worst duplicate exposure on this machine.
+worst duplicate exposure on this machine — and it holds computed either way, over the
+141-file union and over the 171 per-prompt selections with repeats.
+
+**Read the union row, not the sum row.** 171 is the sum of the six per-prompt distinct
+counts, which counts a file once per prompt that selected it; the battery selected
+**141** distinct files in total, so 30 files were picked by more than one prompt. The
+sum row is kept only to make the double-count visible. A later goal that reports a true
+union and compares it to 171 will see a 17% "coverage regression" that never happened.
 
 Definitions, so a later re-measure counts the same things:
 
@@ -329,10 +351,13 @@ Definitions, so a later re-measure counts the same things:
 
 Two observations that are *not* G1 violations but belong on the record:
 
-- **180 context entries cover 171 distinct files.** Nine slots are a second window of
-  a file already selected. That is `MAX_CHUNKS_PER_FILE` working as designed, not
-  duplication — but it means "30 files selected" overstates coverage by ~5%, and any
-  later goal reporting a file count should say which it is counting.
+- **180 context entries cover 141 distinct files.** Battery-wide, 23 of the 180 entries
+  are a second window of a file already selected — `MAX_CHUNKS_PER_FILE` working as
+  designed, not duplication. Within a single prompt the effect is smaller (30 entries →
+  28–29 distinct). So "30 files selected" overstates that prompt's coverage by ~5%, and
+  a raw entry count overstates battery coverage by 28%. Any later goal reporting a file
+  count must say which of the three it is counting: entries, per-prompt distinct, or
+  union.
 - **Named paths were pinned.** P1's `Program.cs` and P2's `EntitlementsController.cs`
   each appear (2 windows each). That is a G2-inv spot check, not a G2-inv
   measurement — the full G2 battery is Goal 3's deliverable.
@@ -343,20 +368,29 @@ Run the M2 battery first (it writes the per-prompt selected-file lists), then:
 
 ```bash
 python3 - <<'PY'
-import subprocess, hashlib, os, collections
+import subprocess, hashlib, os, collections, glob
 REPO = "<platform-root>/m365dotnet"
-for pid in ["P1","P2","P3","P4","P5","P6"]:
-    paths = sorted({l.strip() for l in open(f"/tmp/m2_m365/{pid}_run1.files") if l.strip()})
-    ig = subprocess.run(["git","-C",REPO,"check-ignore","--stdin"],
+
+def g1(label, paths):
+    paths = sorted(set(paths))
+    ig = subprocess.run(["git", "-C", REPO, "check-ignore", "--stdin"],
                         input="\n".join(paths), capture_output=True, text=True)
     ignored = [l for l in ig.stdout.splitlines() if l.strip()]
     h, missing = collections.defaultdict(list), []
     for rel in paths:
-        try: h[hashlib.sha256(open(os.path.join(REPO,rel),'rb').read()).hexdigest()].append(rel)
+        try: h[hashlib.sha256(open(os.path.join(REPO, rel), 'rb').read()).hexdigest()].append(rel)
         except OSError: missing.append(rel)
-    dup = sum(len(v)-1 for v in h.values() if len(v) > 1)
-    wt  = [p for p in paths if p.startswith(".worktrees/") or "/.worktrees/" in p]
-    print(pid, len(paths), len(ignored), dup, len(wt), len(missing))
+    dup = sum(len(v) - 1 for v in h.values() if len(v) > 1)
+    wt = [p for p in paths if p.startswith(".worktrees/") or "/.worktrees/" in p]
+    print(label, len(paths), len(ignored), dup, len(wt), len(missing))
+
+print("label distinct ignored dup in_worktrees unresolvable")
+union = set()
+for pid in ["P1", "P2", "P3", "P4", "P5", "P6"]:
+    p = [l.strip() for l in open(f"/tmp/m2_m365/{pid}_run1.files") if l.strip()]
+    union |= set(p)
+    g1(pid, p)
+g1("UNION", union)   # <- the row the headline quotes
 PY
 ```
 
--- a/docs/eval-baselines-2026-08.md
+++ b/docs/eval-baselines-2026-08.md
@@ -251,10 +251,15 @@ that actually differ: two name a specific file (the pinning path,
 `EXPLICIT_PATH_BOOST`), two are concept-only (pure ranking, no path to pin), one is
 mixed, and one is a symptom/debug phrasing.
 
-| id | shape | prompt |
+The prompt column below is the **verbatim** string the script sends — no backticks, no
+markdown. That is not pedantry: `EXPLICIT_PATH_BOOST` matches on the path token, and a
+stray backtick is part of the prompt text a caller would actually be sending. Copy from
+`tools/m2_battery.sh`, which is the artifact of record; this table is a rendering of it.
+
+| id | shape | prompt (verbatim) |
 |---|---|---|
-| P1 | file-named | Explain what `src/Parslee.M365.Api/Program.cs` does during startup. |
-| P2 | file-named | Add a null check to the entitlements lookup in `src/Parslee.M365.Api/Controllers/EntitlementsController.cs` |
+| P1 | file-named | Explain what src/Parslee.M365.Api/Program.cs does during startup. |
+| P2 | file-named | Add a null check to the entitlements lookup in src/Parslee.M365.Api/Controllers/EntitlementsController.cs |
 | P3 | concept-only | How does the backend authenticate requests from the web app? |
 | P4 | concept-only | Where is chat history persisted and how is it retrieved? |
 | P5 | mixed | Fix the retry logic for Cosmos DB throttling in the repository layer. |
--- a/docs/eval-baselines-2026-08.md
+++ b/docs/eval-baselines-2026-08.md
@@ -1,0 +1,383 @@
+# Neo flagship baselines — August 2026
+
+**Measured:** 2026-08-11
+**Goal:** Unified Store Plan, Goal 1 (Trailhead) — see `docs/unified-store-plan.md`
+**Neo under test:** v0.44.0 at `main` = `77251dff54d0726fb7d98e6e6daa525e1b8b2c6d`
+
+These are the M1 / M2 / G1-invariant baselines every later goal in the unified-store
+plan re-measures against. The mining parameters and the M2 prompt battery below are
+**canonical**: change them and the goal-over-goal comparison is void.
+
+Every number here was produced by running the command printed beside it. Where a
+number could not be produced, the row says so and gives the blocking reason —
+nothing in this document is estimated, projected, or interpolated.
+
+---
+
+## Headline
+
+| Metric | neo | aieweb | m365dotnet | Target |
+|---|---|---|---|---|
+| **M1** MRR (`neo_current`) | 0.136 | 0.180 | **0.051** | m365dotnet ≥ 0.60 |
+| **M1** recall@10 (`neo_current`) | 0.212 | 0.244 | **0.097** | — |
+| **M2** wall-clock, median | — | — | **10.54 s** | ≤ 5 s |
+| **M2** peak RSS (`ru_maxrss`) | — | — | **1.43 GB** | ≤ 500 MB |
+| **G1-inv** gitignored files selected | — | — | **0 / 171** | 0 |
+| **G1-inv** duplicate copies selected | — | — | **0 / 171** | 0 |
+
+Three readings, stated plainly:
+
+1. **Retrieval on the largest flagship is the weakest.** m365dotnet's shipped ranker
+   puts a correct file in the top 10 for roughly one query in ten (recall@10 = 0.097)
+   and its MRR is 0.051 against a 0.60 target — a **11.8×** gap, not a tuning gap.
+2. **The gap is not "retrieval is hard here."** On the same eval set, on the same
+   machine, in the same process, plain BM25 over file content scores MRR 0.607 and
+   recall@10 0.813. The signal is present in the repo; the shipped scorer is not
+   using it.
+3. **G1 is clean post-#186.** The 2026-08-10 reality was 14/16 ignored and 12/16
+   duplicates. Measured now across the whole battery: 0 and 0. This is measured,
+   not assumed.
+
+---
+
+## What was measured, and at which SHA
+
+The goal asked for each repo at `origin/main`. Two of the three could not be moved
+there: `m365dotnet` and `aieweb` both hold uncommitted in-flight work, and this goal
+is forbidden from writing to either. Checking them out to `origin/main` is a write.
+They were therefore measured **in place at their exact current HEAD**, recorded here
+so any later goal can re-measure at the identical tree.
+
+| Repo | Language | Measured at HEAD | Branch | Behind `origin/main` by | Working tree at measurement |
+|---|---|---|---|---|---|
+| neo | Python | `77251dff54d0726fb7d98e6e6daa525e1b8b2c6d` | `main` | 0 | clean (2 untracked: `.worktrees/`, `uv.lock`) |
+| aieweb | TypeScript | `26fff07e0f4a1ec16ca7c9a8446e2bc2d2a4d406` | `feature/platform-nav-customer-tabs-core-rename` | 85 | 39 modified/staged, 14 untracked |
+| m365dotnet | C# | `61dc4a171bdf16cdeb15123dd8790a19248175f9` | `main` | 166 | 1 modified, 3 untracked |
+
+`origin/main` at fetch time, for the record: neo `77251df`, aieweb
+`aa60b64a313b54aed3bf1e11d96e39a809f90f02`, m365dotnet
+`89282099d2ad3e4e474ee22db9c1c7ef1e9f8eee`.
+
+**Consequence to carry forward.** Content is read from the working tree, and the eval
+set is mined from that same HEAD, so each repo is internally consistent — the
+measurement is valid. What it is *not* is a measurement of `origin/main`. A later
+goal re-measuring at a different SHA is comparing two things; re-measure at the SHAs
+above, or re-baseline both sides.
+
+### Machine
+
+| | |
+|---|---|
+| CPU | Apple M4 Max, 16 cores |
+| RAM | 64 GB |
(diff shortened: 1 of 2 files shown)


kkeeling:
## Adversarial Review (local fresh-context reviewer)

Fresh-context, non-author pass on the worktree at `85754a9`. I re-ran the neo and aieweb M1 evals (both reproduce to the last digit), recomputed every M2 aggregate from the raw `/usr/bin/time -l` captures still on disk in `/tmp/m2_m365`, independently re-ran the G1 check over the six per-prompt file lists, recomputed every ratio in the doc by hand, read `tools/m2_battery.sh` line by line, and checked all seven mining-parameter rows against `tools/rank_eval.py`'s actual defaults. **The measurements are sound and the headline table is correct in every cell.** What I found is one internally-contradictory count in the "observations" prose, plus a few smaller accuracy items — notably that the committed battery script has not itself produced the artifacts the doc points at, so its "every number comes out of this one command" contract is slightly ahead of the evidence.

---

### Findings

**1. BLOCKING — "23 of the 180 entries are a second window of a file already selected" is the wrong quantity, and it contradicts the sentence three lines below it.**

`docs/eval-baselines-2026-08.md:354-355`. Measured directly from `/tmp/m2_m365/P*_run1.files`:

```
TOTAL entries 180 | sum-of-per-prompt-distinct 171 | UNION 141
within-prompt repeat entries  = 180 - 171 = 9
battery entries beyond union  = 180 - 141 = 39
raw selected lines: windowed 23, non-windowed 157
```

23 is the number of entries that carry a `(lines a-b)` suffix — i.e. windowed entries, which is exactly the number the footgun paragraph at `:400` correctly reports ("23 of 180 entries carried `" (lines 1-168)"`"). It is *not* the number of entries that are a repeat of an already-selected file. That number is **39** battery-wide, or **9** within-prompt. The doc's own next bullet uses 39 implicitly: "a raw entry count overstates battery coverage by 28%" = (180−141)/141 = 27.7%, which cannot be reconciled with 23 repeats. So the paragraph states two mutually exclusive numbers for the same quantity.

This one is blocking rather than a nit for three reasons: the doc declares itself canonical for every later goal, the sentence sits inside the passage whose entire job is to stop later goals from miscounting, and the PR body's "Update" section promotes the error as a *correction* ("its 'nine slots are a second window' was a per-prompt number presented as battery-wide (23 battery-wide)") — 9 *is* the correct battery-wide within-prompt figure, so that edit replaced a right number with a wrong one and will propagate if merged.

*Fix:* "Battery-wide, 39 of the 180 entries are a repeat of a file already selected — 9 of them a second window within the same prompt (`MAX_CHUNKS_PER_FILE`), the other 30 the same file picked by another prompt." Correct the PR body's update note to match.

**2. NON-BLOCKING — "so 30 files were picked by more than one prompt" overstates the file count.**

`docs/eval-baselines-2026-08.md:334`. 171 − 141 = 30 is the *excess selection count*, not a file count; a file picked by three prompts contributes 2. Measured: **21** distinct files were picked by more than one prompt, with a total excess of 30. The surrounding claims are unaffected — 171, 141 and the "17% coverage regression" ((171−141)/171 = 17.5%) are all correct — but this clause is the same label-vs-quantity slip as finding 1, in the paragraph telling later goals to be precise about which count they are quoting.

*Fix:* "…so the six prompts made 30 more selections than there are distinct files (21 files were picked by more than one prompt)."

**3. NON-BLOCKING — the PR body still says G1 was checked "over all 171 distinct selected files".**

Under *How Has This Been Tested?*. 171 is not a distinct count — the body's own Update section says so, and the doc says so twice. The correct figure is 141. The doc is right; only the body is stale.

**4. NON-BLOCKING — the committed `tools/m2_battery.sh` did not produce the artifacts the doc reports, and its CSV will disagree with the doc's `context entries` column when someone runs it.**

Evidence: the timed captures `/tmp/m2_m365/P*_run*.txt` are stamped 17:19–17:21; `tools/m2_battery.sh` is stamped 17:41; `results.csv` and `union.files` are stamped 17:41. Running the script rewrites the `.txt` files, so the committed script cannot have produced them. Concretely, `results.csv` carries `files_selected = 30` for every row, but the committed pipeline computes that column as `sort -u` over the *suffix-stripped* paths — replaying it on `P1_run1.txt` yields **29**, not 30:

```
grep -E '^  .+ - [0-9]+ bytes' P1_run1.txt \
  | sed -E 's/^  (.+) - [0-9]+ bytes.*/\1/; s/ \(lines [0-9]+-[0-9]+\)$//' | sort -u | wc -l   # 29
```

So a reader who runs the reproduction command gets a `files_selected` column of 29/28/29/29/28/28, while the doc's M2 table prints `context entries` 30 for all six — two different quantities under two labels that look interchangeable, which is the precise confusion the doc warns about at `:358`. The published wall-clock/RSS numbers are unaffected (I verified all 18 of them against the raw captures — see below), and the *aggregate* block does print entries/sum/union correctly. But the doc's claim at `:297` that "every number in the M2 tables comes out of this one command" is not established for that column.

*Fix:* either rename the CSV column to `distinct_files` and add an `entries` column, or drop the per-prompt `context entries` column from the M2 table and let the G1 table own it. A single end-to-end run of the committed script would also close the gap between the artifact set and the script.

**5. NIT — `<platform-root>` is used in four commands but never defined, and the doc uses three placeholder spellings.**

`:177`, `:218`, `:223`, `:292`, `:372` use `<platform-root>`; `:215`/`:291` use `<neo-worktree>`; `:194` uses `<neo>`; `:413` uses `<repo>`. Using placeholders instead of a developer-specific absolute path is the right call and I am not asking for hardcoded paths — but nothing tells a reader that `<platform-root>` is the parslee-knowledge root that the flagship repos hang off. One line ("`<platform-root>` = the directory containing `neo/`, `aieweb/`, `m365dotnet/`") makes all four commands runnable, and standardising `<neo>` → `<neo-worktree>` removes the third spelling.

**6. NIT — `tools/m2_battery.sh` runs with `set -uo pipefail` but no `-e`, so a failed `neo` invocation is recorded as an empty field and surfaces later as a `ValueError`.**

`tools/m2_battery.sh:25,58-59`. If `/usr/bin/time -l "$NEO_BIN" …` fails, `awk` yields empty `wall`/`rss`, the row is written as `P1,file-named,1,,`, and the Python aggregation dies on `float("")` *after* `results.csv` has already been tee'd — a battery that looks half-successful. Related: the aggregation globs `P*_run1.files` in `$OUT`, so re-running into a directory left over from a different prompt set silently folds stale files into the union. Both are cheap to close (check `$wall`/`$rss` non-empty and fail loud; `rm -f "$OUT"/P*_run*.files` before the loop).

**7. NIT — the doc's battery table renders the P1/P2 paths in backticks; the script's prompt strings have none.**

`docs/eval-baselines-2026-08.md:245-246` vs `tools/m2_battery.sh:36-37`. Markdown formatting, not a measurement difference — but the table is labelled canonical and someone re-keying from it would send the backticks into a prompt whose whole purpose is exercising `EXPLICIT_PATH_BOOST`. Since the script is the artifact of record, a one-line note ("prompts shown formatted; the literal strings are in `tools/m2_battery.sh`") settles it.

---

### What I verified, and it held

- **Scope.** `git diff --stat main...HEAD` = exactly two added files, `docs/eval-baselines-2026-08.md` (+417) and `tools/m2_battery.sh` (+114). No `src/`, no `tests/`. Mode `100755` on the script, `#!/usr/bin/env bash` present, so the printed reproduction command runs as written.
- **M1 neo reproduces exactly.** `rank_eval.py --eval <neo> /tmp/cases_neo.json` → 209 evaluable of 210; `neo_current` 0.043 / 0.098 / 0.146 / 0.212 / 0.418 / MRR 0.136; `bm25_content` MRR 0.559. Every cell matches the doc.
- **M1 aieweb reproduces exactly.** 221 of 225; `neo_current` 0.053 / 0.117 / 0.182 / 0.244 / 0.296 / MRR 0.180; `bm25_content` MRR 0.726. Every cell matches.
- **Eval-set sizes.** From the mined case files: neo 214 → 210 after the `release ` filter; aieweb 225 → 225; m365dotnet 173 → 173. Matches the table.
- **All 18 M2 timings.** Re-derived from the raw `/usr/bin/time -l` captures. Every per-prompt median/min/max in the results table is right (P1 10.52/10.44/11.43 … P6 11.26/11.06/11.97), and every per-prompt peak RSS converts correctly (P2's 1,434,402,816 B → 1367.95 MiB → prints 1368.0).
- **Battery median 10.54 s follows from the 18 runs.** Sorted, the 9th/10th values are 10.52 and 10.55 → 10.535 → `f"{:.2f}"` → 10.54. min 7.45 and max 11.97 are the true extremes. Peak RSS 1368.0 MiB is the max of the column.
- **Every ratio.** 0.60 ÷ 0.051 = 11.76 → "11.8×" ✓. 1,434,402,816 ÷ 500e6 = 2.869 → "2.87×" ✓, and the MiB alternative is 2.74× understating by 4.6% ≈ "5%" ✓. 10.535 ÷ 5 = 2.11 → "2.1×" ✓. (180−141)/141 = 27.7% → "28%" ✓. 30 ÷ 28.5 = 1.053 → "~5%" ✓. (171−141)/171 = 17.5% → "17%" ✓.
- **G1 re-run independently.** Piping each prompt's paths through `git -C m365dotnet check-ignore --stdin` plus SHA-256 content grouping reproduces the table row for row: P1 29, P2 28, P3 29, P4 29, P5 28, P6 28, UNION **141** — and **0 ignored, 0 duplicates, 0 in `.worktrees/`, 0 unresolvable** at every row. The invariant claim holds. 180 entries, 171 sum-of-rows, 141 union are all confirmed.
- **The "not measured" claims are honest.** `tools/rank_eval.py:114-119` sets `has_idx = bool(idx.chunks)`; `:147-148` sets `dn = []` when false; `:151` is `"rrf_bm25_dense": rrf([bmr, dn]) if dn else bmr`. Both descriptions in the doc are literally accurate, and my two eval runs confirm it empirically — `dense` printed 0.000 across the board and `rrf_bm25_dense` came back byte-identical to `bm25_content` on both repos. `ProjectIndex.__init__` does set `self.neo_dir = self.repo_root / ".neo"` at `src/neo/index/project_index.py:157`, exactly as cited, so the "index build writes inside the measured repo" blocker is real.
- **Every mining-parameter row.** `max_commits=400` ✓, `--no-merges` ✓, `1 <= len(files) <= 3` ✓, `len(subject) > 15` ✓, the nine extensions ✓, the test-exclusion regex `(^|/)(tests?|spec)/|(^|/)test_|_test\.` ✓, the case-insensitive `release ` drop applied at `--eval` ✓, `ks=(1,3,5,10,20)` ✓. All seven rows are the actual defaults. The CLI exposes exactly `--build-from-git` and `--eval REPO CASES` as printed, and `--build-from-git` writes JSON to stdout with the count on stderr, so the `> /tmp/cases_$r.json` redirect is clean.
- **The battery script does what the doc says.** 1 discarded warm-up (`:47`) + `RUNS=3` × 6 prompts = 18 timed runs ✓; median/peak/ratios derived in the embedded Python, not by hand ✓; writes `<id>_run<n>.txt`, `<id>_run<n>.files`, `results.csv`, `union.files` ✓; strips the `(lines a-b)` suffix ✓; the six prompt strings match the doc's table (modulo finding 7's backticks) ✓. The `grep`/`sed` pattern matches `cli.py:1105`'s literal output format, and the dry-run listing goes to **stderr**, which is what the script captures — so the `2>"$raw"` redirect is correct, not an accident.
- **`shellcheck tools/m2_battery.sh`** — clean, exit 0. `ruff check tools/` — all checks passed. Both match the PR body's claims.
- **No leakage.** m365dotnet is still at `61dc4a17…` and aieweb at `26fff07e…`, matching the recorded HEADs; neither has a `.neo/` directory. m365dotnet does carry **44** in-repo worktrees, as claimed.
- **The pinning spot-check is real.** `P1_run1.txt` shows `src/Parslee.M365.Api/Program.cs` at lines 512-591 and 6562-6653, both `score: 10.00`; `P2_run1.txt` shows `EntitlementsController.cs` twice at `score: 12.45`. Two windows each, as stated.

I did not execute the M2 battery (expensive, and the raw captures made it unnecessary), and I did not re-run the m365dotnet M1 eval (275 s) — that one rests on the artifact set plus the fact that neo and aieweb both reproduced exactly.

VERDICT: CHANGES REQUESTED — 1 blocking finding.

kkeeling:
## Author response — every finding, fixed or declined

Reviewer was right on the blocking finding. Per-finding disposition below; fixes are in `d64d47a` and `96c65b5`.

| # | Finding | Severity | Disposition |
|---|---|---|---|
| 1 | `23` attached to the wrong quantity | BLOCKING | **Fixed** `d64d47a` |
| 2 | "30 files picked by >1 prompt" | NON-BLOCKING | **Fixed** `d64d47a` — 21 files / 30 excess selections |
| 3 | PR body says G1 checked over "171 distinct" | NON-BLOCKING | **Fixed** — body now says 141 |
| 4 | Committed script never produced the reported artifacts; `files_selected` reads 29 vs the doc's 30 | NON-BLOCKING | **Fixed** `d64d47a` — script re-run end-to-end, two columns now |
| 5 | Undefined `<platform-root>` placeholder | NIT | **Fixed** `d64d47a` |
| 6 | Missing `set -e` | NIT | **Declined as written; addressed differently** — see below |
| 7 | Backticked prompts in the doc table | NIT | **Fixed** `96c65b5` |

### 1 — the blocking finding

Confirmed by recomputation from the raw artifacts, not by agreement. `23` is the count of *windowed* entries, which is the number the footgun paragraph cites correctly. The repeat-entry counts are different quantities:

| quantity | value | derivation |
|---|---|---|
| context entries | 180 | 6 × 30 |
| sum of per-prompt distinct | 171 | 29+28+29+29+28+28 |
| **battery union distinct** | **141** | the number to quote |
| repeat entries, battery-wide | **39** | 180 − 141 |
| repeat entries, within-prompt | **9** | 180 − 171 |
| files selected by >1 prompt | **21** | (30 excess selections) |

You also caught that the round-1 Update in the PR body promoted this as a *correction* — it replaced 9, which was the right within-prompt figure, with 23. That claim is struck in the body. **No measurement changes: G1-inv is still 0 ignored / 0 duplicates over the 141-file union.**

The script now derives all three repeat quantities and prints them, so the number this fixes comes out of a command rather than out of prose.

### 4 — the reproduction contract, now actually proven

You were pointing at a real gap: the published tables came from an ad-hoc first pass and `tools/m2_battery.sh` was written afterwards, so the *committed* script had never been shown to produce the published numbers. It has now been run end-to-end into a clean directory (3 min 28 s):

| quantity | first pass | committed-script re-run |
|---|---|---|
| entries / per-prompt distinct / **union** | 180 / 171 / **141** | 180 / 171 / **141** |
| per-prompt distinct P1…P6 | 29, 28, 29, 29, 28, 28 | 29, 28, 29, 29, 28, 28 |
| the union file set itself | — | **byte-identical** |
| repeat entries battery-wide / within-prompt | 39 / 9 | 39 / 9 |
| G1-inv (ignored / dup / `.worktrees/` / unresolvable) | 0 / 0 / 0 / 0 | 0 / 0 / 0 / 0 |
| peak `ru_maxrss` | 1368.0 MiB (2.87×) | 1368.0 MiB (2.87×) |
| battery median wall | **10.54 s** | 10.22 s |

Selection is deterministic; wall-clock is not. The whole 3% wall move is one scheduler stall — P4 run 1 at 19.27 s against 10.18 s on run 3 of the same prompt — which is exactly why the protocol reports a median. **The published baseline stays 10.54 s**; re-baselining to the second sample would be picking a number for no reason. The doc now states the ±0.5 s noise band so a later goal does not read noise as a result.

Your `files_selected` observation was correct and is why `results.csv` now emits `entries` **and** `distinct_files`. One ambiguously-named column is how a 30 in one table gets compared against a 29 in another.

### 6 — `set -e` declined as written

`grep -c` returns 1 on zero matches, so `set -e` would abort the battery mid-run on a legitimate-looking condition, and `pipefail` is already set. The underlying concern — a silently failed run writing an empty CSV cell that the aggregation reads as a number — is real, so it is guarded explicitly instead: the loop aborts loudly if `wall`, `rss`, or the entry count is empty/zero.

One thing that guard needed and did not have at first: the measurement loop is the left side of a `| tee`, so it runs in a **subshell** and its `exit 1` would have been swallowed — it would have printed `FATAL` and then aggregated the partial CSV anyway. The pipeline status is now checked after the `tee`.

### 7 — worth more than "nit"

The battery is declared **canonical** for every later goal in the plan, so the doc's rendering and the script's `PROMPTS` array disagreeing is a correctness problem, not cosmetics: `EXPLICIT_PATH_BOOST` matches on the path token, and anyone re-keying a prompt from the doc would have sent a different string than the one that produced these numbers. All six rows are now verified byte-identical to the array.

### Gates

`shellcheck tools/m2_battery.sh` clean · `ruff check src/ tests/ tools/` all checks passed · scope unchanged (`docs/` + `tools/` only).

kkeeling:
## Adversarial Re-Review (local fresh-context reviewer) — head 96c65b5

Fresh-context, non-author pass on `96c65b5`. I did not write this code and did not write the round-1 review. I recomputed all six disputed counts from the raw artifacts in `/tmp/m2_m365` (first pass) **and** `/tmp/m2_m365_verify` (the committed-script re-run), executed the script's Python aggregation block against the re-run CSV, exercised the new FATAL guard end-to-end with a stub binary, re-ran the G1 check independently against m365dotnet, re-ran the neo M1 eval, and swept the doc for every occurrence of every count. **All six numbers are right, the blocking finding is genuinely fixed, and the fix commits introduce no defect.** The one thing I found in the new code is an unreachable edge case in the guard's numeric test, reported below as a nit with its own disproof of reachability.

---

### Prior findings

| # | Finding | Severity | Status | Evidence at this head |
|---|---|---|---|---|
| 1 | `23` attached to the repeat-entry quantity | BLOCKING | **RESOLVED** | `docs/eval-baselines-2026-08.md:399-410` now states **39** battery-wide (180−141) and **9** within-prompt (180−171), each with its derivation, and explicitly ring-fences **23** as the *windowed*-entry count. Recomputed from raw: 39, 9, 23 — all three match. I also checked the wording, not just the arithmetic: all 9 within-prompt repeats really are second windows (`all_repeats_windowed=True` for each of P1–P6), so "9 … a second window of a file already selected *by that same prompt*" is literally true, not merely numerically true. The 23 decomposes cleanly as 9 files × 2 windows + 5 single-window files. |
| 2 | "30 files picked by >1 prompt" | NON-BLOCKING | **RESOLVED** | `:378-379` — "30 *excess selections* spread over **21 files**". Recomputed: 21 files with count > 1, excess 171−141 = 30. |
| 3 | PR body says G1 checked over "171 distinct" | NON-BLOCKING | **RESOLVED** | Body now reads "over all **141** distinct selected files (the battery union)". |
| 4 | Committed script never produced the reported artifacts; `files_selected` 29 vs doc's 30 | NON-BLOCKING | **RESOLVED** | `/tmp/m2_m365_verify/` exists, stamped 21:55–21:58, one `.txt`/`.files` per run plus `results.csv`. CSV header is now `id,shape,run,wall_seconds,max_rss_bytes,entries,distinct_files` and rows read `30,29` — the two quantities are now separately named. The union set is **byte-identical** across the two passes (`a == b → True`, 141 == 141), confirming the determinism claim rather than taking it. |
| 5 | Undefined `<platform-root>` | NIT | **RESOLVED** | `:15-24` defines both `<platform-root>` and `<neo-worktree>`. |
| 6 | No `set -e`; empty CSV cell → later `ValueError` | NIT | **PARTIALLY** | The stated concern is fixed and I verified it end-to-end (see below). The *related* half of that finding — `rm -f "$OUT"/P*_run*.files` before the loop, so a re-run into a dirty directory can't fold stale files into the union glob at `:129` — was not addressed and is not mentioned in the response. Still a nit: with the canonical battery the files are overwritten, so it only bites a re-run with a *different* prompt set. |
| 7 | Backticked prompts in the doc table | NIT | **RESOLVED** | Verified independently by parsing the `PROMPTS` array out of the script and string-comparing against the doc table cells: all six **exact-match**, including P2's 105-character prompt. |

### New findings from the fix commits

**N1 — NIT (unreachable). `[ "$nentries" -eq 0 ]` is not empty-string-safe under bash.** `tools/m2_battery.sh:78`. If `nentries` were empty, bash prints `[: : integer expression expected`, the test returns 2, and the guard **falls through** rather than firing:

```
$ bash -c 'set -uo pipefail; wall=1; rss=2; nentries=""; if [ -z "$wall" ] || [ -z "$rss" ] || [ "$nentries" -eq 0 ]; then echo GUARD_FIRES; else echo FELL_THROUGH; fi'
bash: line 0: [: : integer expression expected
FELL_THROUGH
```

I am reporting this as a nit rather than a defect because I could not reach it, and the reachability argument cuts the right way: `$OUT/${id}_run${run}.files` is unconditionally created by the redirect at `:72`, so `grep -c .` at `:73` always has a readable file and always emits a number — `0` on an empty file, verified — which is precisely the case `|| true` exists for and which the guard handles correctly. And if it *were* reached, the blast radius is a blank CSV cell, not the `ValueError` the original finding warned about, because the Python block reads only `wall_seconds` and `max_rss_bytes`. Nothing to change unless you want belt-and-suspenders (`[ "${nentries:-0}" -eq 0 ]`).

No other new findings. Specifically checked and clean:

- **The `grep -c … || true` idiom.** `grep -c` exits 1 on zero matches but still prints `0`, so the substitution captures `0` and the guard fires. Under `pipefail`, `sort -u | grep -c .` at `:74` behaves the same.
- **The FATAL guard and the post-`tee` `status=$?`.** Exercised end-to-end with a stub `neo` that emits no selected lines. The subshell's `exit 1` is surfaced by `pipefail`, caught at `:90-94`, and the script exits **1** with the aggregation never running — no partial aggregate, no crash. Comments between the pipeline and `status=$?` do not clobber `$?`.
- **The CSV rename vs `csv.DictReader`.** The aggregation reads `id`, `shape`, `wall_seconds`, `max_rss_bytes` only. `files_selected` was never read, so renaming it to `entries` + `distinct_files` cannot break it — confirmed by execution, not inspection.
- **`collections` is imported** at `:105` and used for `defaultdict` (`:108`) and the new `Counter` (`:128`).

### Verified at this head

- **The script's aggregation block executes clean** against `/tmp/m2_m365_verify/results.csv` (exit 0), printing `180 entries | 171 sum-of-per-prompt-distinct | 141 BATTERY UNION`, `repeat entries: 39 battery-wide | 9 within-prompt`, `files selected by >1 prompt: 21 (accounting for 30 excess selections)`, battery median **10.22 s**, peak **1368.0 MiB**, **2.87×** memory — every figure the new "Reproducibility re-run" section publishes.
- **All six required counts, recomputed from raw artifacts in both directories:** 180 entries · 171 sum-of-per-prompt-distinct · 141 union · 39 battery-wide repeats · 9 within-prompt repeats · 21 files picked by >1 prompt · 30 excess selections. Identical in both passes. Per-prompt distinct 29/28/29/29/28/28.
- **Every M2 cell re-derived from the raw `/usr/bin/time -l` captures**, independent of `results.csv`: P1 10.52/10.44/11.43/1367.8 … P6 11.26/11.06/11.97/1366.9; battery median **10.54** (n=18, min 7.45, max 11.97); peak 1,434,402,816 B = **1368.0 MiB = 1.43 GB**. Ratios: 2.11 → "2.1×", 2.8688 → "**2.87×**", MiB base 2.74× understating by 4.6% → "5%". All correct.
- **G1 re-run independently** via the doc's own reproduction snippet against `m365dotnet` at `61dc4a17…`: P1–P6 and UNION all `ignored=0 dup=0 in_worktrees=0 unresolvable=0`, union **141**, in both directories. The invariant holds.
- **Doc internal consistency.** Swept every occurrence of 171, 141, 39, 9, 21, 30, 23. Every `171` is labelled "sum of rows" / "sum of the six per-prompt distinct counts" — none is presented as a distinct count. Every `141` is the union. Derived percentages check out: 39/141 = 27.7% → "28%", 30/28.5 = 5.3% → "~5%", 30/171 = 17.5% → "17%", (10.54−10.22)/10.54 = 3.0% → "±3%".
- **Prompt table is byte-identical to `PROMPTS`** — all six, verified by programmatic comparison.
- **M1 neo reproduces exactly**: 209 of 210 evaluable, `neo_current` 0.043/0.098/0.146/0.212/0.418/MRR **0.136**, `bm25_content` MRR **0.559** — matching `:151-152`. The run also confirms the "not measured" edit is honest: `dense` printed 0.000 across the board and `rrf_bm25_dense` came back byte-identical to `bm25_content`, exactly as `tools/rank_eval.py:150-151` (`rrf([bmr, dn]) if dn else bmr`, with `dn = []` when `bool(idx.chunks)` is false) predicts.
- **Spot checks.** P1's `Program.cs` at lines 512-591 and 6562-6653 (both score 10.00) and P2's `EntitlementsController.cs` at 1-106 and 1455-1566 (both 12.45) — 2 windows each, as `:411-413` says. `.worktrees/` holds **44** entries, matching `:357`. No `.neo/` in m365dotnet; HEAD unmoved at `61dc4a17…`.
- **Gates.** `bash -n tools/m2_battery.sh` clean · `shellcheck tools/m2_battery.sh` exit 0, no output.
- **Scope.** `git diff --stat main...96c65b5` = `docs/eval-baselines-2026-08.md` (+467) and `tools/m2_battery.sh` (+144), 2 files, 611 insertions. No `src/`, no `tests/`.
- **CI.** `gh pr checks 201` — 4 jobs (`test` on 3.10/3.11/3.12/3.13), all **pending** at the time of writing (run 31555485878). Not a review finding; it is a merge gate the maintainer should see green first.

I did not re-run the full M2 battery (unnecessary — the committed artifacts of both passes were on disk and I recomputed from them) and did not re-run the aieweb or m365dotnet M1 evals.

VERDICT: APPROVE — no blocking findings.

kkeeling:
## Both re-review nits taken — head `46caa36`

The re-review returned **APPROVE, 0 blocking** ([comment](https://github.com/Parslee-ai/neo/pull/201#issuecomment-5261251443)) and raised two nits. Both are real and both are now fixed, because both are the exact class of error this document exists to make impossible.

**1 — stale artifacts in a reused output directory.** The aggregation globs `"$OUT"/P*_run1.files` to build the battery union, so a re-run into a dirty directory with a changed prompt set would fold a previous run's leftovers into the union and report a distinct count that no single battery ever produced. The script now clears its own artifacts first. The `rm` pattern is pinned to the exact names it writes — `P[0-9]_run[0-9].{files,txt}`, plus `union.files` and `results.csv` — so pointing `OUT` at a populated directory cannot delete anything the script did not write itself.

**2 — `[ "$nentries" -eq 0 ]` on an empty string.** Correct: under bash that errors and falls *through* rather than firing, which would let the very row the guard exists to catch reach the CSV. Unreachable today — the `.files` file is always created by the redirect, so `grep -c .` always emits a number — but the guard should not depend on that staying true. It is `[ "${nentries:-0}" -eq 0 ]` now.

Verified the guard end-to-end against a stub binary that measures nothing:

```
id,shape,run,wall_seconds,max_rss_bytes,entries,distinct_files
FATAL: P1 run 1 produced no measurement (wall='0.01' rss='1933312' entries=0); see /tmp/m2_guard/P1_run1.txt
FATAL: measurement loop aborted (exit 1); /tmp/m2_guard/results.csv is incomplete.
exit=1
```

FATAL prints, the post-`tee` status check catches the subshell exit, the script exits 1, and the aggregation never runs on the partial CSV. Re-ran the aggregation block against the real `/tmp/m2_m365_verify/results.csv` afterwards to confirm no regression — still `180 entries | 171 sum-of-per-prompt-distinct | 141 BATTERY UNION`, `39 / 9` repeats, `21` files, `2.87x` memory.

`bash -n` clean · `shellcheck` clean · scope unchanged (`docs/` + `tools/` only).

No numbers in the document changed. **G1-inv remains 0 gitignored / 0 duplicate copies over the 141-file union.**

kkeeling:
## Adversarial Re-Review (local fresh-context reviewer) — head 46caa36

Delta check only, on `96c65b5..46caa36`. The commit touches one file (`tools/m2_battery.sh`, +13/−1) and contains exactly the two changes it claims. **The delta is clean; my round-2 approval carries to this head.**

### The `rm -f` — the one worth attacking

`tools/m2_battery.sh:39-40`. I tried to make it delete something it should not, and could not:

| attack | result |
|---|---|
| `OUT` at a populated dir with decoys (`P1_run1.files.bak`, `P1_run1.txt.orig`, `P_run1.files`, `Q1_run1.files`, `PROGRAM.txt`, `important.txt`, `subdir/P1_run1.files`) | only the four exact-name patterns died; **every decoy survived**, and it is non-recursive so the subdir copy was untouched |
| `OUT` unset **or** empty | unreachable — `OUT="${3:?}"` at `:29` uses `:?`, which fails on unset *or null*, so `rm -f /P[0-9]…` cannot be formed (verified both cases abort) |
| `OUT` containing a glob metacharacter (`/tmp/rm*star`) | quoted `"$OUT"` stays literal, so only the literally-named dir was touched; a sibling `/tmp/rmESCAPE/P1_run1.files` survived |
| `OUT` containing spaces | correct — artifact removed, `keep.me` survived |
| no matches at all | `rm -f` exits 0, silent (bash passes the unexpanded pattern; `-f` swallows it) |

The quoting is right: `"$OUT"` quoted, pattern unquoted, which is the only arrangement that both handles spaces/metacharacters in `OUT` and still expands the pattern inside it. **Not blocking, nothing to change.**

**Downstream:** nothing breaks. `results.csv` is regenerated by the `tee` and `union.files` by the Python block in the same run; the doc's G1 snippet reads `P*_run1.files`, also regenerated. Nothing is expected to persist across runs. A *failed* re-run now destroys the prior good artifacts — but that was already true at `96c65b5` (the `tee` truncated `results.csv` to its header, and the `.txt`/`.files` were overwritten in place), so it is not a regression.

### `${nentries:-0}`

Fixes exactly the fall-through I reported: `:-` substitutes on unset **or null**, so empty → `0` → the guard fires. It cannot coerce a real measurement, because a real `nentries` is a non-empty numeric string and `:-` never touches it — the only new behaviour is that an empty value now fails loud instead of silently reaching the CSV. That is the fail-safe direction: it can produce a false FATAL, never a false pass. Re-verified end-to-end with a stub `neo` at this head — FATAL printed, script exit **1**, `results.csv` header-only, aggregation never ran.

### New finding

**NIT (non-blocking, narrower than the gap it closes).** The `rm` pattern is `P[0-9]_run[0-9]` but the aggregation globs `P*_run1.files` at `:138`, so the two disagree on ids outside a single digit. Measured: `P10_run1.files` and `P_run1.files` both **survive the `rm` and are seen by the aggregation glob**. Reaching it requires editing the canonical `PROMPTS` array to double-digit ids *and* re-running into a dirty directory — i.e. a narrower case of the scenario the `rm` was added for, and strictly better than the previous state of no cleanup at all. (`RUNS>9` similarly leaves `P1_run10.txt` behind, but those never match the `_run1.files` union glob, so that one is clutter only, not a wrong number.) Fix if you want symmetry: `P*_run*.files` / `P*_run*.txt`.

### Verified at this head

- `bash -n` clean · `shellcheck` exit 0, no output · worktree file byte-identical to committed `46caa36`, `git status` clean.
- Aggregation heredoc **byte-identical** to `96c65b5` (diffed), and re-executed against `/tmp/m2_m365_verify/results.csv`: exit 0, still prints `180 entries | 171 sum-of-per-prompt-distinct | 141 BATTERY UNION`, `39` / `9` repeats, `21` files / `30` excess, median 10.22 s, 1368.0 MiB, 2.87×. No regression.
- Scope: `git diff --stat main...46caa36` = `docs/eval-baselines-2026-08.md` (+467) and `tools/m2_battery.sh` (+156). Still docs/ + tools/ only, no `src/`, no `tests/`.
- No measurement, count, or doc number is touched by this commit.

VERDICT: APPROVE — no blocking findings.

kkeeling:
## Status — ready for maintainer merge, head `46caa36`

| gate | result |
|---|---|
| CI (`test` 3.10 / 3.11 / 3.12 / 3.13) | **4/4 SUCCESS** on `46caa36` |
| Adversarial review, round 1 (head `85754a9`) | CHANGES REQUESTED — 1 blocking, resolved |
| Adversarial review, round 2 (head `96c65b5`) | **APPROVE** — 0 blocking |
| Delta re-review (head `46caa36`) | **APPROVE** — 0 blocking |
| Scope | `docs/` + `tools/` only, 2 files, no runtime or product code |

Nine substantive items were raised across the three rounds and nine were taken. One is deliberately **deferred**, recorded here rather than fixed:

**Deferred nit — `rm` pattern narrower than the aggregation glob.** The cleanup matches `P[0-9]_run[0-9]`, while the union is built from `P*_run1.files`, so a hypothetical `P10_run1.files` would survive the `rm` and still be seen by the glob. Reaching it requires editing the `PROMPTS` array past nine prompts *and* reusing a dirty output directory. The reviewer's own assessment is that the narrow pattern is strictly better than no cleanup, and its narrowness is what makes it provably unable to delete anything the script did not write — which it was attacked five ways to confirm. Widening it to match the glob trades that proof for an unreachable case. Left as-is on purpose; if a later goal extends the canonical battery past nine prompts, that is the moment to revisit, and this comment is the pointer.

### Merge authority

`#200` merged at 23:23Z, putting the ruling on `main`: *"goals open PRs; auto-merge on green gates is authorized (Keenan, 2026-08-11)"* (`docs/unified-store-plan.md:18`). This goal's constraint makes merge authorized once CI is green with the numbers pasted — both now hold.

**This PR is not merged by its author.** `neo` is absent from `~/.car-pr-review/repos.json`, so no Grace verdict will ever arrive here and the local fresh-context reviewer above stands in for it — but an agent merging its own work is the thing that review exists to prevent. Leaving it for @kkeeling.

### Final numbers (unchanged by any review round)

| Metric | neo | aieweb | m365dotnet | Target |
|---|---|---|---|---|
| **M1** MRR (`neo_current`) | 0.136 | 0.180 | **0.051** | m365dotnet ≥ 0.60 |
| **M1** recall@10 (`neo_current`) | 0.212 | 0.244 | **0.097** | — |
| **M2** wall-clock, median | — | — | **10.54 s** | ≤ 5 s |
| **M2** peak RSS (`ru_maxrss`) | — | — | **1.43 GB** | ≤ 500 MB |
| **G1-inv** gitignored files selected | — | — | **0 / 141** | 0 |
| **G1-inv** duplicate copies selected | — | — | **0 / 141** | 0 |

Every review round corrected *reporting*, never a measurement. The G1 invariant has read 0 / 0 over the 141-file union at every head, and the committed battery script reproduces the union byte-identically.

kkeeling:
## Branch brought up to date — head `6481566`, all automated gates green

`main`'s branch protection sets `strict: true` (require branches up to date before merging), so being one commit behind was a genuine merge blocker rather than cosmetic drift. `origin/main` is merged in — a clean `ort` merge bringing only `docs/unified-store-plan.md`, the plan of record this PR measures against, which now resolves in-tree.

Scope is unchanged after the merge: `git diff --name-only origin/main...HEAD` is still exactly `docs/eval-baselines-2026-08.md` and `tools/m2_battery.sh`. No number in the document moved.

| gate | state on `6481566` |
|---|---|
| CI `test` 3.10 / 3.11 / 3.12 / 3.13 | **4/4 SUCCESS** |
| behind `main` | **0** |
| `mergeable` | MERGEABLE |
| scope | `docs/` + `tools/` only |
| adversarial review | APPROVE, 0 blocking (rounds 2 and 3) |
| **remaining** | **`reviewDecision = REVIEW_REQUIRED` — 1 approving review** |

That last row is the only thing left, and it is deliberately not mine to clear. The local fresh-context reviewer stands in for the Grace daemon because `neo` is absent from `~/.car-pr-review/repos.json`, but it posts comments, not GitHub approvals — and an agent approving or merging its own PR is precisely what the review gate exists to prevent. Over to @kkeeling for the approving review and merge.

PR #202, opened 2026-08-12
test(release-gate): per-language LLM round trip gates the release; invariant battery gates every PR

kkeeling:
## Description

Wires Neo's release process to a per-language LLM round trip, and every pull request to the free half of the same battery.

`build` in `publish.yml` now needs a new `language-roundtrip` job: one real `neo --json` call each for **C#, TypeScript and Python** against generated fixture repositories. A red language stops the chain at `build` and nothing reaches PyPI. `ci.yml` gains an `invariants` job that runs the same fixtures with **no model call** on every push and PR.

This is Goal 10 (G5-inv) of [`docs/unified-store-plan.md`](https://github.com/Parslee-ai/neo/blob/main/docs/unified-store-plan.md). It touches tests, CI and docs only — no gatherer or index runtime code (Goals 5–8 own those files).

## Type of Change

- [x] Documentation update

## Related Issue

Guards the class of failure behind #158 / #159 (C# absent from the index for 8.5 months) and #186 (gatherer selected gitignored junk). Does not close them.

## Motivation and Context

Matt Liotta, on the 2026-08-10 sync: real-LLM round-trip tests per language belong in the release process — *"if that's what we have to do as part of the release process, we want to do it because we obviously want to catch that shit."* Cost-bearing model calls are approved **for releases**, not for every push.

Three failures motivate the language axis specifically, and all three were invisible from the outside:

- **C# was absent from Neo's index for 8.5 months.** Everything that could have caught it was exercised against Python, and passed.
- **The gatherer selected gitignored junk** — 14 of 16 selected files on one measured prompt.
- **Every failure printed success and exited 0.**

A single-language check would have stayed green through all of it.

### The two halves

| | Free half | Paid half |
|---|---|---|
| File | `tests/test_selection_invariants.py` | `tests/test_release_roundtrip.py` |
| Marker | `-m invariants` | `-m roundtrip` |
| Model calls | none | one per language |
| Runs on | every push and PR | the release flow, and manual dispatch |
| Wired in | `invariants` job, `ci.yml` | `language-roundtrip` job, `publish.yml` |
| Blocks | the PR | the wheel — `build` needs it |

Both build the **same** three fixture repos from `tests/fixtures/language_repos.py`. That sharing is deliberate: a red release gate has to be diagnosable from a free CI run, and it cannot be if the two halves disagree about the repository under test.

Each generated fixture is a real git repository (the invariants are defined against `git check-ignore`, and a fake cannot be differentially compared to git) holding, in its own language: a **target** file the prompt names, carrying a **sentinel** symbol present in no other file; **gitignored junk** in the same language sharing the target's filename stem and sentinel, so it competes for the same slots; a **duplicate** of the target under `.claude/worktrees/…` which is *not* gitignored, so the gatherer's own defaults must catch it; and **bulk below the sentinel** so the target exceeds the prompt's per-file cap and the truncation assertions are non-vacuous on the file that matters.

The load-bearing round-trip assertion is that **the model names the sentinel**. It exists in exactly one file, so it cannot be named unless that file reached the prompt — closing the gap between "the gatherer selected it" and "the model saw it", which is the gap C# sat inside.

### 1. Full round-trip suite, all three languages green (local, real LLM)

`NEO_INFERENCE_MODE=auto NEO_RELEASE_ROUNDTRIP=1 pytest -m roundtrip -v`

```
test_the_result_is_not_empty[csharp] PASSED
test_the_result_is_not_empty[python] PASSED
test_the_result_is_not_empty[typescript] PASSED
test_the_model_names_the_symbol_only_that_file_contains[csharp] PASSED
test_the_model_names_the_symbol_only_that_file_contains[python] PASSED
test_the_model_names_the_symbol_only_that_file_contains[typescript] PASSED
test_confidence_is_reported[csharp] PASSED
test_confidence_is_reported[python] PASSED
test_confidence_is_reported[typescript] PASSED
test_the_orchestrator_envelope_is_present[csharp] PASSED
test_the_orchestrator_envelope_is_present[python] PASSED
test_the_orchestrator_envelope_is_present[typescript] PASSED
test_the_named_file_is_selected_and_whole[csharp] PASSED
test_the_named_file_is_selected_and_whole[python] PASSED
test_the_named_file_is_selected_and_whole[typescript] PASSED
test_nothing_gitignored_is_selected[csharp] PASSED
test_nothing_gitignored_is_selected[python] PASSED
test_nothing_gitignored_is_selected[typescript] PASSED
test_no_duplicate_copies_are_selected[csharp] PASSED
test_no_duplicate_copies_are_selected[python] PASSED
test_no_duplicate_copies_are_selected[typescript] PASSED

21 passed, 2195 deselected in 123.33s (0:02:03)
```

### 2. Invariant battery (free, no model call)

`pytest -m invariants` → **48 passed, 2168 deselected in 29.71s**. The `invariants` CI job on this PR is the run the DoD asks for.

### 3. Mutation check — the battery actually fails when the product breaks

Each mutation applied to `src/`, battery run, then reverted (`git checkout --`; working tree confirmed clean after).

| Mutation | Result |
|---|---|
| Gatherer stops reading the repo's `.gitignore` | **6 failed**, 42 passed |
| Agent-worktree layout exclusions removed | **6 failed**, 42 passed |
| Prompt renderer cuts without a marker | **3 failed**, 45 passed |

### 4. The paid half stays opted out

`pytest -m roundtrip` with no env var → **21 skipped**, each with the reason printed. They skip rather than deselect on purpose: an ordinary run should *show* that the gate exists. The `invariants` CI job asserts this negative directly — it fails if any round trip executes on a PR, and fails if none were collected at all. Three tests in the battery pin the gate itself (env-gated, marker present, all three languages in `LANGUAGES`), so deleting it turns every PR red rather than expensive.

### 5. Credential-absent behaviour

Reproduced locally: with no provider configured, the round trip fails with `Failed to initialize LM adapter`. The CI job asserts the secret up front and **fails rather than skips**. An absent credential is a red gate, not a green one.

**Test Configuration**: Python 3.13.7 · macOS (darwin 25.5.0) · neo @ this branch · provider routed via CAR locally, Anthropic in CI.

## Checklist

- [x] New and existing unit tests pass locally with my changes — see the two caveats below

## Additional Notes

Three things a reviewer should decide on rather than discover.

**1. `ANTHROPIC_API_KEY` is not configured on this repo, so the gate blocks releases until it is added.** That is the gate working, and it is deliberate — a job that skips when unconfigured is the same defect as a failure that exits 0. It needs adding at Settings → Secrets and variables → Actions before the next release. Optional variable `NEO_RELEASE_MODEL` overrides the model; the default matches `adapters.create_adapter`'s own Anthropic default so the gate and the library cannot drift.

**2. The round trip can go red for reasons that are not about file selection, and it is not retried.** Measured during bring-up: two consecutive local runs of the same commit, one 21/21 green, the next red on Python with `ValidationError: Failed to parse simulation traces: missing_start_sentinel`. The gatherer had selected the right four files both times; the model's reply simply lacked neo's v3 sentinel. A user issuing that command gets the same error, so it is a genuine release blocker — and a gate that retries until green is the defect this goal exists to prevent. What the PR does instead is make the failure *name its layer*, and record that `pytest -m invariants` will be **green** when the parser is the cause. That contrast is the fastest diagnosis available and it costs nothing. Expect it to bite occasionally; the right response is to fix the brittleness, not loosen the gate.

**3. One pre-existing test failure on `main`, untouched by this branch.** `tests/test_integration.py::TestPerformanceIntegration::test_consolidation_performance_with_all_boosts` asserts a wall-clock `<100ms` bound and measures **198.2ms on this machine**. Reproduced on `origin/main` @ `5bbee46` with an identical `src/`; this branch changes no file under `src/` and does not touch `test_integration.py`. Left alone as out of scope. Worth noting it is the same defect class `CLAUDE.md` already documents for the learning-loop benchmark — a hardware-dependent latency threshold gating a correctness verdict — and it will flap on shared runners.

**Also included, one line:** `publish.yml`'s `ci-check` was linting `src/ tests/` while `ci.yml`, the Makefile and the pre-commit hook lint `src/ tests/ tools/`. The release path enforced less than a pull request did. Now aligned.

Full reference: [`docs/release-gate.md`](https://github.com/Parslee-ai/neo/blob/goalpool/g_mspyjqfj_ddd931-unified-store-goal-10/docs/release-gate.md).

### DIFF
--- a/.claude/commands/ship-release.md
+++ b/.claude/commands/ship-release.md
@@ -97,6 +97,19 @@ git push origin v{version}
 
 ### Phase 6: Create GitHub Release
 
+**The release gate runs here.** Creating the release triggers `publish.yml`,
+which will not build a wheel until `language-roundtrip` is green: one real LLM
+round trip each for C#, TypeScript and Python against generated fixture
+repositories, asserting that the language's files reach the prompt and that
+the model can name a symbol present in exactly one of them. A red language
+stops the chain at `build` and nothing reaches PyPI.
+
+The job **fails** rather than skips when `ANTHROPIC_API_KEY` is not configured
+as a repository secret — an absent credential is a red gate, not a green one.
+Configure it before the first gated release. Full details, including how to
+dry-run the gate without cutting a release, are in
+[`docs/release-gate.md`](../../docs/release-gate.md).
+
 ```bash
 gh release create v{version} \
   --title "v{version}" \
@@ -121,6 +134,7 @@ byte-identical artifacts, so a mismatch would not be visible on the release page
 ### Phase 7: Verify Publication
 
 Check that:
+- `language-roundtrip` passed — all three languages green
 - GitHub Actions workflow completed successfully
 - Package appears on PyPI: https://pypi.org/project/neo-reasoner/
 
@@ -157,6 +171,13 @@ above `publish-testpypi` in the workflow, and CONTRIBUTING.md.
 **If tag already exists**: Report conflict, suggest incrementing version
 
 **If GitHub Actions fails**: Check workflow logs at github.com/{repo}/actions
+
+**If `language-roundtrip` fails**: reproduce it for free first —
+`pytest -m invariants -v` runs the same fixtures with no model call and will
+usually name the broken invariant. If the battery is green and only the round
+trip is red, the model stopped seeing a file the gatherer still selects; run
+`NEO_RELEASE_ROUNDTRIP=1 pytest -m roundtrip -v` locally. Do not work around
+it by removing the job from `build`'s `needs`.
 
 **If PyPI publish fails**: Check Actions logs for authentication or build issues
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,3 +30,47 @@ jobs:
 
     - name: Run tests
       run: pytest -v
+
+  # The free half of the release gate, as its own required-looking check.
+  #
+  # These assertions also run inside `test` above — this job exists so that a
+  # broken selection invariant is legible AS a broken selection invariant on
+  # the PR that caused it, rather than as one red line inside a 2,000-test
+  # matrix. That legibility is the whole point: C# was absent from Neo's index
+  # for 8.5 months (#158/#159) and the gatherer selected gitignored junk
+  # (#186), and in both cases something green was being read as evidence.
+  #
+  # No model call, no credentials, every PR. The paid per-language round trip
+  # is release-only and lives in publish.yml; see docs/release-gate.md.
+  invariants:
+    name: Guard-invariant battery
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+
+    steps:
+    - uses: actions/checkout@v7
+
+    - name: Set up Python
+      uses: actions/setup-python@v7
+      with:
+        python-version: '3.11'
+
+    - name: Install dependencies
+      run: pip install -e ".[dev]"
+
+    - name: Guard invariants (G1-inv..G3-inv, per language, no LLM)
+      run: pytest -m invariants -v
+
+    # Proves the negative the constraint actually cares about: the paid half
+    # was collected and did NOT run. A gate that silently stopped being
+    # release-only would otherwise first show up on the bill.
+    - name: Confirm the LLM round trip stayed opted out
+      run: |
+        pytest -m roundtrip -v --no-header -rs > roundtrip-run.txt 2>&1
+        cat roundtrip-run.txt
+        if grep -qE '[0-9]+ (passed|failed)' roundtrip-run.txt; then
+          echo "::error::round-trip tests executed on a PR; they are release-only"
+          exit 1
+        fi
+        grep -qE '[0-9]+ skipped' roundtrip-run.txt || {
+          echo "::error::no round-trip tests were collected at all"; exit 1; }
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -23,14 +23,80 @@ jobs:
       - name: Install dependencies
         run: pip install -e ".[dev]"
 
+      # Same three paths as ci.yml, the Makefile and .githooks/pre-commit.
+      # This job had been linting two of them, so the release path enforced
+      # LESS than a pull request did — a release could publish code that CI
+      # would have rejected.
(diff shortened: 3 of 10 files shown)


kkeeling:
## CI evidence (2026-08-12)

### 1. Invariant battery executing on this PR

All five checks green on [run 31591434856](https://github.com/Parslee-ai/neo/actions/runs/31591434856):

| Check | Result | Time |
|---|---|---|
| **Guard-invariant battery** (new) | pass | 47s |
| test (3.10) | pass | 4m29s |
| test (3.11) | pass | 4m16s |
| test (3.12) | pass | 4m30s |
| test (3.13) | pass | 4m26s |

The new job's `Confirm the LLM round trip stayed opted out` step proves the negative — the paid half was collected and did not run:

```
SKIPPED [3] tests/test_release_roundtrip.py:173: release-only: costs a real LLM call per language. Set NEO_RELEASE_ROUNDTRIP=1 to run (see docs/release-gate.md).
... (6 more, one per assertion class)
22 skipped, 2194 deselected in 2.05s
```

No `passed`, so the step's guard is satisfied. (The 22nd skip is `test_car_host_smoke.py` skipping at import via `importorskip("car_runtime")` — a collection-time skip, not a marker selection. `-m roundtrip` selects exactly the 21 gate tests.)

The `test` matrix also passed on all four Python versions, including `test_consolidation_performance_with_all_boosts` — the wall-clock test noted as failing locally at 198ms. It passes on GitHub runners. That is the flap, not a fix.

### 2. Release-flow dry run — the gate fires

`gh workflow run publish.yml --ref goalpool/g_mspyjqfj_ddd931-unified-store-goal-10` → [run 31591442379](https://github.com/Parslee-ai/neo/actions/runs/31591442379), conclusion **failure**:

```
ci-check:                                    success
Per-language LLM round trip (release gate):  failure
build:                                       skipped
publish-pypi:                                skipped
publish-testpypi:                            skipped
```

Failed at the `Require release credentials` step, which is the designed behaviour with no `ANTHROPIC_API_KEY` secret configured. **`build` and both publish jobs skipped. Nothing reached PyPI.**

That is the DoD's dry run: the gate would fire, and a red gate stops the chain before a wheel is ever built. It also makes the outstanding action concrete — add the `ANTHROPIC_API_KEY` repository secret before the next release, or this is the failure every release will hit.

PR #204, opened 2026-08-12
feat(index): give ruby and php the edge queries they never had (re-lands #190)

kkeeling:
## Description

Re-lands **#190** (`feat(index): give ruby and php the edge queries they never had`), which GitHub reports as MERGED but whose changes are **not on `main`**. Same two commits by Jennifer Liotta, rebased onto `main`; the `language_parser.py` diff is byte-identical to what #190 merged, verified before opening this PR.

Ruby declared **no edge queries at all** — the only language in `QUERIES` absent from `EDGE_QUERIES`. Every ruby file in every index contributed nodes to the graph and never a relationship: no `require`, no superclass, no mixin. PHP declared imports only, so `class Impl extends BaseImpl implements Greeter` produced nothing where the java, c_sharp and typescript equivalents each produce edges. Both gaps were found while writing the per-query yield tests in #187 and pinned there as failing-when-fixed assertions.

The second commit is the one worth reading: the first cut used `(base_clause (name) @base)`, which requires `name` as a **direct** child, while the real shape is `base_clause → qualified_name → name` (Ruby's is `superclass → scope_resolution → constant`). `class ApplicationRecord < ActiveRecord::Base` — the single most common inheritance statement in the Ruby ecosystem — produced **nothing**, and `class B extends \Foo\Bar implements \ArrayAccess, Countable` produced exactly one edge, for `Countable`. Silent *partial* extraction, which is worse than a total miss because the consumer sees a plausible answer with no signal that most of it is absent. It is also precisely what the `_CS_BASE_TYPES` comment 150 lines above warns about.

## Type of Change

## Related Issue

Re-lands #190. Follow-on from #158. Goal 2 of `docs/unified-store-plan.md` — Land the queue.

### Why #190 says MERGED while `main` does not have it

Recorded in full because the state is genuinely confusing to anyone reading the PR list, and because the mechanism will recur on any stacked PR:

1. #190's base was **`fix/dry-run-shows-what-it-claims`** (#187's branch), not `main`.
2. At `2026-08-12T12:33:11Z` #190 was squash-merged **into that branch** as `54a24df`. Its content reached #187's branch and never `main`.
3. Goal 2 had checked #187's branch out at `64e1ee1` — the tip *before* that merge — and rebased it onto `main` for the #194 landing.
4. Force-pushing the rebased #187 replaced `54a24df`, taking #190's commit off the branch with it. #187 then squash-merged to `main` carrying its own five commits and none of #190's.
5. Deleting #187's branch on merge left #190 pointing at a base that no longer exists, so GitHub closed it as MERGED-into-a-deleted-branch. It cannot be reopened (`can't be reopened because it was already merged`) and its base cannot be retargeted (`Cannot change the base branch of a closed pull request`).

So a new PR is the only route, and this is not a duplicate of an open one. Confirmed against `main` rather than inferred: `origin/main:src/neo/index/language_parser.py` contains one occurrence of `'ruby'` where this branch has two, and the file is 111 lines short.

**The lesson, for the next stacked pair:** rebase and land the *bottom* of a stack only after re-fetching, and do not `--delete-branch` while another PR is based on it. Either land the stack bottom-up in one pass, or retarget the upper PR to `main` **before** merging the lower one.

## How Has This Been Tested?

- [x] `pytest tests/test_language_query_yield.py tests/test_selection_invariants.py` — **90 passed** in 128 s, including #202's new guard-invariant battery
- [x] Full `pytest` suite — results posted as a comment below
- [x] `ruff check src/ tests/ tools/` — all checks passed
- [x] The `#match?` predicate that separates `require 'x'` from `puts 'hello'` is enforced by its own test, since a tree-sitter that stopped applying predicates would keep compiling and start emitting an import edge per string-argument call — silent, and in the noisiest direction

**Test Configuration**:
- Python version: 3.13.7 (CI additionally covers 3.10–3.12)
- OS: macOS 26.5.2
- Neo version: 0.44.0 (`main` at `8abfb34`)

## Checklist

- [x] Any dependent changes have been merged and published — #194 (`281cf87`) and #187 (`8abfb34`) are both on `main`

## Additional Notes

Authorship is Jennifer Liotta's on both commits; the rebase preserved it. Nothing in the diff was authored by the rebasing agent.

### DIFF
--- a/src/neo/index/language_parser.py
+++ b/src/neo/index/language_parser.py
@@ -303,6 +303,62 @@ _CS_INHERITANCE = "[\n" + "\n".join(
 ) + "\n] @class_def"
 
 # Edge queries for relationship extraction (imports, inheritance)
+# PHP supertypes reach the query in two shapes and BOTH must be covered:
+#
+#     implements Countable          -> (name)
+#     extends \Foo\Bar             -> (qualified_name (name))   (everyday PSR)
+#
+# The narrow arm alone compiles perfectly and drops every namespaced
+# supertype without a word -- `class B extends \Foo\Bar implements
+# \ArrayAccess, Countable` yielded exactly one edge, `B implements
+# Countable`, which is worse than a total miss because the consumer has no
+# signal anything is absent. This is the same failure the `_CS_BASE_TYPES`
+# comment above documents, reproduced in a new language 150 lines below it.
+#
+# Do NOT copy the C# spelling `(qualified_name name: (name) @x)` -- PHP's
+# `qualified_name` has no `name:` field and that pattern fails to compile.
+# The unfielded direct child is the last segment, which matches C#'s
+# rightmost-segment convention.
+_PHP_BASE_TYPES = "[(name) @{cap} (qualified_name (name) @{cap})]"
+
+# Generated across declaration kinds rather than written out per kind, for
+# the reason `_CS_INHERITANCE` is generated: a hand-maintained duplicate
+# invites the next widening to be applied to some of the arms and not the
+# rest. PHP 8.1 enums carry `class_interface_clause` too, and were missed by
+# the hand-written version.
+_PHP_EXTENDS_KINDS = ("class_declaration", "interface_declaration")
+_PHP_IMPLEMENTS_KINDS = ("class_declaration", "enum_declaration")
+
+
+def _php_edge_queries() -> dict:
+    base = _PHP_BASE_TYPES.format(cap="base")
+    iface = _PHP_BASE_TYPES.format(cap="interface")
+    extends = " ".join(
+        f"({kind} name: (name) @class_name (base_clause {base}))"
+        for kind in _PHP_EXTENDS_KINDS
+    )
+    implements = " ".join(
+        f"({kind} name: (name) @class_name (class_interface_clause {iface}))"
+        for kind in _PHP_IMPLEMENTS_KINDS
+    )
+    # `use SomeTrait;` inside a class body. Mapped to `implements` for parity
+    # with Ruby's mixins in this same change: both mean "this type gains that
+    # module's surface", and two languages in one commit taking opposite
+    # stances on the same construct would not survive the next maintainer.
+    traits = (
+        f"(class_declaration name: (name) @class_name "
+        f"(declaration_list (use_declaration {iface})))"
+    )
+    return {
+        'imports': """
+            (namespace_use_declaration
+                (namespace_use_clause (qualified_name) @module)) @import
+        """,
+        'inheritance': f"[ {extends} ] @class_def",
+        'implements': f"[ {implements} {traits} ] @class_def",
+    }
+
+
 EDGE_QUERIES = {
     'python': {
         'imports': """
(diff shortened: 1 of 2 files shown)


kkeeling:
## Test results

**CI — all green** (`31599640391`, head `8ffce65`):

| check | result | time |
|---|---|---|
| test (3.10) | pass | 6m40s |
| test (3.11) | pass | 6m32s |
| test (3.12) | pass | 7m23s |
| test (3.13) | pass | 7m34s |
| Guard-invariant battery (#202) | pass | 48s |

3.11 failed on the first attempt and was re-run. The failure was `test_learning_evaluation.py::TestLatencyIsNotACorrectnessGate::test_within_budget_run_reports_clean` — `latency_ms=687.63 exceeds latency_ms_max=500.0`, with `accepted=True` and `acceptance_failures=[]` on the same report. That is a wall-clock reading on a shared runner, on a benchmark this repo has already litigated (#183/#185: an 11× spread between a laptop and a runner on identical code), and it is untouchable by a change to ruby and php tree-sitter queries. Re-ran green; every other version passed first time.

**Local — targeted:** `pytest tests/test_language_query_yield.py tests/test_selection_invariants.py` → **90 passed** in 128 s. That is this PR's own per-query yield suite plus #202's guard-invariant battery, on Python 3.13.7 / macOS 26.5.2.

**Local — full suite:** still running at time of merge and materially slower than CI on this box (the same suite took 17 min, then 37 min, then 49+ min across today's three runs while the tree did not change in any way that touches runtime). Since CI executes the identical suite on four interpreters and does so in seven minutes, CI is the gate here; the local run is a duplicate, not a stronger signal. Result appended below when it lands.

kkeeling:
### Addendum — local full suite, as promised above

`pytest` on this box finished at **2370 passed, 29 skipped, 6 failed in 4227 s (1h 10m)**. CI ran the same suite in 6–7 minutes per interpreter and passed on all four, so the honest reading is that this machine is degraded, not that the tree is.

All six failures are the same two shapes, and none of them can be reached by a change to ruby and php tree-sitter queries:

| test | shape |
|---|---|
| `test_integration.py::…::test_consolidation_performance_with_all_boosts` | wall-clock assertion (`elapsed_ms < 100`) |
| `test_learning_evaluation.py::TestLatencyIsNotACorrectnessGate::test_within_budget_run_reports_clean` | wall-clock assertion (`latency_ms_max=500`) |
| `test_cli_global_flags.py::…::test_no_scan_dry_run_does_not_require_api_key` | subprocess timeout |
| `test_cli_index_flag.py::test_index_flag_basic_functionality` | subprocess timeout |
| `test_learning_evaluation.py::test_evaluation_cli_runs_without_provider_keys_or_user_memory` | subprocess timeout |
| `test_operating_modes.py::…::test_verify_cli_needs_no_provider_and_agent_cli_fails_before_provider` | subprocess timeout |

The subprocess tests all run `python -m neo` under `HOME=tmp_path` against a 30 s budget, so a cold fastembed cache is re-fetched inside the budget on every one of them. Two of these were re-run in isolation against `main` **with this branch's changes absent** and failed identically there, which is what rules the branch out.

Worth recording separately, because it is a slow-moving signal rather than a flake: the identical suite on this machine took **17 min, then 37 min, then 70 min** across today's three runs, on trees that differ only in tree-sitter queries and test files. Whatever that is — accumulated `~/.neo` state, thermal, or the model cache — it is not something a PR caused, and it is why the four wall-clock assertions above are worth revisiting as the #183/#185 lesson says: a latency threshold is not a correctness gate, and there are still four of them in this suite.

PR #206, opened 2026-08-12
fix(trust): language-aware constraint markers (#196) and a confidence scale without a sentinel (#199)

### LINKED ISSUE #196: constraint_verifier reports 'no obvious handler' unconditionally on every non-Python codebase
`CONSTRAINT_CODE_MARKERS` in `src/neo/constraint_verification.py:89-96` maps every constraint type to **Python-only** source tokens:

```python
CONSTRAINT_CODE_MARKERS: Dict[ConstraintType, tuple] = {
    ConstraintType.SORTED:          ("sorted(", ".sort(", "heappush", "heappop", "bisect"),
    ConstraintType.INCREASING:      ("sorted(", ".sort(", "bisect"),
    ConstraintType.DECREASING:      ("sorted(", ".sort(", "reverse=True"),
    ConstraintType.UNIQUE_ELEMENTS: ("set(", "dict.fromkeys"),
    ConstraintType.NON_NEGATIVE:    ("abs(", "max(0"),
    ConstraintType.DIVISIBILITY:    ("%",),
}
```

Five of the six can never match C#, TypeScript, Java, Go, or C++. `sorted(` / `.sort(` / `heappush` / `bisect` / `set(` / `dict.fromkeys` are Python spellings; the C# equivalents are `OrderBy`, `.Sort()`, `HashSet<T>`, `Math.Abs`, `Math.Max`. Only `DIVISIBILITY` (`%`) is language-agnostic.

The static checker in `engine.py:2478` then emits "generated code shows no obvious handler (expected one of: ...)" — **unconditionally**, for any non-Python target that trips one of these constraint patterns.

## Observed

2026-08-11, reasoning about a C# codebase (Parslee ODI, .NET 10). Neo returned:

```
Still open: constraint_verifier found 1 issues
  Prompt declares constraint 'Elements must be unique' but generated code shows
  no obvious handler (expected one of: set(, dict.fromkeys)
```

The target in that run was a **markdown** file (`/decision.md`), so the checker fired a Python-handler check against prose. The constraint itself was never at issue.

## Why it matters beyond noise

The comment above the table says "Absence of a marker is a warning, not an error — the LM may satisfy the constraint through other means." That reasoning covers a *different approach*; it does not cover a *different language*, where the marker set cannot match by construction.

The practical effect is a permanently-firing caution on every non-Python repo. That is alarm fatigue on the same channel that carries Neo's real warnings — including the genuinely valuable ones, e.g. "I ran out of time before the checkers. This code is unverified," and Neo's correct refusals to answer on partial context. A caution channel that always has an entry trains operators to skip it.

## When the fix lands

A C# file satisfying a uniqueness constraint via `HashSet<T>` or `.Distinct()` produces no constraint_verifier caution. A test asserts that the marker check either resolves per-language or is skipped when the target language is not Python — rather than reporting an unsatisfiable expectation.

Options, no strong preference: key the marker table by detected language; skip the check when the language is unknown or non-Python; or replace token matching with a language-aware query (the repo already has per-language parsing under `src/neo/index/language_parser.py`).

## Scope

No open PR touches `src/neo/constraint_verification.py` as of filing (checked #186, #187, #190, #191, #192, #194). Confirmed present on `origin/main`.

### LINKED ISSUE #199: Confidence 0.5 is a 'no suggestions' sentinel, so a correct analysis scores lower than an empty patch
`_calculate_confidence` in `src/neo/engine.py:2528`:

```python
if not suggestions:
    return 0.5

avg_confidence = sum(s.confidence for s in suggestions) / len(suggestions)
...
return max(0.0, min(1.0, avg_confidence - issue_penalty - check_penalty))
```

Overall confidence is the mean of the model's **self-reported per-suggestion** confidences, and `0.5` when there are no suggestions at all. Two consequences, both observed:

## 1. `0.5` conflates "not a code-change task" with "moderately uncertain"

A pure analysis/review run legitimately produces zero `CodeSuggestion`s. It reports `0.5` regardless of how well-grounded the answer is. That value is not a measurement — it is a sentinel — but nothing downstream distinguishes it from a genuine mid-confidence result.

## 2. Declining to answer can score higher than answering correctly

Because the average is over self-reported values and an empty diff still carries its self-reported number, a run that produces no actual content can outscore one that does.

## Observed, 2026-08-11 (same session, same subject: a C# refresh/equivalence audit)

| Run | Output | Confidence |
|---|---|---|
| Planner-stage, `unified_diff: ""`, explicitly declined to name findings for lack of context | **no content** | **0.96** -> `READY_TO_IMPLEMENT` |
| Full analysis, three sub-questions answered, every verdict verified against source afterwards | **correct content** | **0.50** -> `Proceed with caution` |
| Earlier run, *same verdicts* as the 0.50 run | correct content | **0.92** |

The 0.96 run's own `code_suggestions[0].description` read: "No speculative patch is provided at planner stage because the supplied source excerpts omit portions ... needed to answer the property audit accurately." That is correct, honest behaviour — and it scored `READY_TO_IMPLEMENT` per `subcommands.py:1455`, while the run that actually answered scored "Proceed with caution - some uncertainties remain" (`subcommands.py:1509`).

The 0.92 vs 0.50 pair is the same answer twice at a 42-point spread, decided by whether the run happened to emit a suggestion object rather than by anything about the answer.

Note the 0.96 suggestion also had a real `file_path` with an empty `unified_diff`, which `engine.py:3815` states is only permitted when `file_path` is `/` or `N/A` — so a schema-violating suggestion contributed full weight to the score.

## Why this is more than cosmetic

Confidence is consumed as if calibrated:

- `EARLY_EXIT_CONFIDENCE = 0.8` (`engine.py:172`) gates skipping static checks (`engine.py:1168`, `engine.py:2686`)
- `confidence_interpretation` maps ranges to operator-facing actions (`subcommands.py:1455`, `1508-1509`)

So the number that decides "skip the checkers" and "ready to implement" is, for analysis runs, a constant — and for empty-patch runs, the model's opinion of a patch it did not write.

## When the fix lands

- A run with no code suggestions reports absent/not-applicable confidence rather than `0.5`, and `confidence_interpretation` renders that distinctly instead of "some uncertainties remain".
- A suggestion with an empty `unified_diff` and a real `file_path` is either rejected per the existing schema rule or excluded from the average.
- A test asserts that an analysis-only run and a genuine mid-confidence run are distinguishable in the output.

Separately worth considering: the current value is self-report averaged, with no term for whether anything was verified. The same session's runs carried cautions like "I ran out of time before the checkers. This code is unverified" and `verification_verdict: skipped` while scoring 0.95+ — the score does not know the checks did not run.

kkeeling:
## Description

Two trust-calibration defects, both of which made a Neo channel say the same thing
regardless of the run.

**#196** — `CONSTRAINT_CODE_MARKERS` was a single Python-only table applied to every
target, so the "no obvious handler (expected one of: `set(`, `dict.fromkeys`)" caution
could not be cleared by any C#, TypeScript or markdown code. Markers are now keyed by
language; a language with no table produces an honest `info` note naming the language
instead of a warning, and every downstream counter reads actionable diagnostics only.

**#199** — `confidence = 0.5` was a "no suggestions" sentinel sitting on the very scale
the number is read against, so a run that emitted an empty patch with a self-reported
`0.96` outranked a run that answered correctly at `0.50`. The sentinel comes off the
scale, and a suggestion naming a real file with no diff and no code block is excluded
from the average and from the early-exit gate.

While writing the #196 tests, a **second, larger half of the same defect** surfaced:
`_strip_comments_and_strings` rebuilds the source by joining tokens with a space, so
`sorted(x)` reached the matcher as `sorted ( x )` and the marker `sorted(` could never
match. The caution therefore fired on satisfied **Python** code too — see the BEFORE
block below, where all five cases warn. Both sides are now whitespace-normalized
before matching, in the fail-safe direction (a false match only suppresses a warning;
absence of a marker was always a hint, not a verdict).

Plan of record: `docs/unified-store-plan.md`, "Goal 4. Trust calibration". Opened as a
**DRAFT** per the 2026-08-12 Review & merge governance ruling — the shepherd flips it
ready after evidence verification.

## Type of Change

- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code cleanup/refactoring

## Related Issue

Fixes #196
Fixes #199

## Motivation and Context

A caution channel with a permanent entry teaches operators to skip the channel — the
same channel that carries "I ran out of time before the checkers. This code is
unverified" and Neo's correct refusals. And a confidence number that is a constant for
analysis runs, and the model's opinion of a patch it did not write for empty-patch
runs, is consumed as if it were calibrated: `EARLY_EXIT_CONFIDENCE = 0.8` decides
whether the checkers run at all, and `confidence_interpretation` renders operator-facing
actions off it.

### BEFORE — `origin/main` @ `4a48b80`

```
#196 — constraint_verifier
  C# satisfying uniqueness via HashSet<T>    -> [warning] ... no obvious handler (expected one of: set(, dict.fromkeys)
  C# with no handler                         -> [warning] ... no obvious handler (expected one of: set(, dict.fromkeys)
  markdown target (the reported run)         -> [warning] ... no obvious handler (expected one of: set(, dict.fromkeys)
  Python satisfying uniqueness               -> [warning] ... no obvious handler (expected one of: set(, dict.fromkeys)
  Python with no handler                     -> [warning] ... no obvious handler (expected one of: set(, dict.fromkeys)

#199 — confidence
  empty patch, real path, self-reported 0.96   -> confidence=0.96   action=READY_TO_IMPLEMENT
  correct analysis, no suggestions             -> confidence=0.50   action=PROCEED_WITH_CAUTION
  genuine mid-confidence change                -> confidence=0.55   action=PROCEED_WITH_CAUTION
```

Five out of five constraint cases warn — the check is a constant. And the empty patch
outranks the correct analysis by 46 points and two action tiers.

### AFTER — this branch

```
#196 — constraint_verifier
  C# satisfying uniqueness via HashSet<T>    -> no caution
  C# with no handler                         -> [warning] Prompt declares constraint 'Elements must be unique' but generated csharp code shows no obvious handler (expected one of: HashSet<, ToHashSet(, .Distinct(, .GroupBy(, ISet<)
  markdown target (the reported run)         -> [info] Constraint 'Elements must be unique' was not checked: no markdown marker set for constraint type 'unique_elements'
  Python satisfying uniqueness               -> no caution
  Python with no handler                     -> [warning] Prompt declares constraint 'Elements must be unique' but generated python code shows no obvious handler (expected one of: set(, dict.fromkeys)

#199 — confidence
  empty patch, real path, self-reported 0.96   -> confidence=None   basis=no_verifiable_change   action=NO_VERIFIABLE_CHANGE
  correct analysis, no suggestions             -> confidence=None   basis=analysis_only          action=ANALYSIS_ONLY
  genuine mid-confidence change                -> confidence=0.55   basis=suggestions            action=PROCEED_WITH_CAUTION
```

The #196 permanent false alarm is gone on the C# fixture while the real missing-handler
case still warns on Python. The #199 inversion is corrected: the correct analysis now
ranks **above** the empty patch, and the mid-confidence run stays distinguishable from
both.

Stated precisely, because the first draft of this line overclaimed: the inversion is
fixed by the values themselves — neither run reports a number any more, and the two
report *different* bases and different operator-facing actions, so there is no longer a
scale on which the empty patch can outrank the analysis. `models.CONFIDENCE_ACTIONS` /
`confidence_action_rank` **declare** that ordering for hosts and tests; they are not
read by any internal control flow, and calling them "the ladder the fix runs on" would
describe a table nobody executes.

Both blocks were produced by the same script against the two source trees; reproduce by
pointing `PYTHONPATH` at each `src/` and calling `_check_constraints_static` /
`_calculate_confidence` with the fixtures pinned in the two new test modules.

## What changed

- **`constraint_verification.py`** — `LANGUAGE_CONSTRAINT_MARKERS` keyed by language
  (Python, C#, TypeScript/JavaScript), `language_for_path`, `markers_for_language`.
  `CONSTRAINT_CODE_MARKERS` stays as an alias for the Python table, which is what the
  name always meant.
- **`engine.py`** — `_check_constraints_static` resolves markers per language and emits
  `info` "not checked for `<language>`" where none exists; `_normalize_for_marker_match`
  fixes the token-rejoin miss; `_is_actionable_diagnostic` / `_actionable_diagnostics`
  are read by issue counts, next-questions and the confidence penalty;
  `_static_check_status` returns `skipped` (not `passed`) for an info-only result, so a
  check that evaluated nothing cannot vouch for an early exit.
  `_calculate_confidence` returns `(Optional[float], basis)`; the early-exit gate and
  the average both filter to scoreable suggestions.
- **`models.py`** — `suggestion_is_scoreable` / `suggestion_is_review_marker`, the four
  `CONFIDENCE_BASIS_*` constants, `NeoOutput.confidence: Optional[float]` plus
  `confidence_basis`, and the `CONFIDENCE_ACTIONS` trust ladder with
  `confidence_action_rank`.
- **`subcommands.py`** — `_interpret_confidence` renders `ANALYSIS_ONLY` and
  `NO_VERIFIABLE_CHANGE` distinctly instead of folding them onto the numeric bands; the
  scale reference now names the `null` case. Numeric bands are unchanged.
- **`cli.py` / `car_tool_schema.py`** — `confidence_basis` on both wire shapes;
  `CONFIDENCE: n/a (<reason>)` instead of printing `0.50`.
- **`config/beats/neo_matrix.yaml`** — the three new user-facing lines, per the
  no-prose-in-engine rule.

## How Has This Been Tested?

- [x] `tests/test_constraint_language_markers.py` — 25 tests: the reported markdown run,
      C#/TypeScript satisfaction and misses in each language's own vocabulary, the
      Python regression guard, the token-rejoin regression, honest degradation for
      unmapped extensions, and that info notes raise no next-question, dock no
      confidence and do not read as `passed`.
- [x] `tests/test_confidence_calibration.py` — 16 tests: the #199 pair reproduced
      end-to-end through `engine.process`, plus the mechanisms and the wire contract.
- [x] Guard-invariant battery: `pytest -m invariants` — **48 passed, 1 skipped**.
- [x] `ruff check src/ tests/ tools/` — clean.
- [x] Affected-surface sweep (orchestrator events, CAR tool schema, JSON serialization,
      operating modes, subcommands, engine, multi-agent, static analysis, schemas,
      structured parser, host-adapter parity, import health, home isolation):
      **315 passed, 1 failed**.

**The one local failure is pre-existing and environmental**, verified by running the
same test against `origin/main` @ `4a48b80` in the same shell:
`test_operating_modes.py::test_verify_cli_needs_no_provider_and_agent_cli_fails_before_provider`
gives the CLI subprocess 30 s, and the subprocess spends ~20 s of it fetching the
fastembed model into the test's fresh `HOME` on a machine under load 35. It fails
identically on `origin/main`. Running the same VERIFY payload by hand on this branch
returns `rc=0` with `confidence_basis: "verification_verdict"`. CI is the authoritative
full-suite verdict.

**Test Configuration**:
- Python version: 3.13.7 (CI matrix: 3.10 / 3.11 / 3.12 / 3.13)
- OS: macOS (darwin 25.5.0)
- Neo version: 0.45.0

## M1 — branch vs main, three flagships

No selection code is touched. The diff is 7 source files
(`car_tool_schema.py`, `cli.py`, `neo_matrix.yaml`, `constraint_verification.py`,
`engine.py`, `models.py`, `subcommands.py`) and 2 new test modules; nothing under
`context_gatherer`, `index/` or the walker. `--dry-run` returns before any `NeoOutput`
is built, so it never reaches the changed `cli.py` block or `_calculate_confidence` at
all.

M1 is computed from what `neo --dry-run` selects, so identical selection means identical
R@k / MRR for any ground truth. Measured directly — same prompts, same repos, same venv,
only `PYTHONPATH` differing between this branch's `src/` and `origin/main` @ `4a48b80`:

```
neo         q1: IDENTICAL  (25 selected entries, branch == main)
neo         q2: IDENTICAL  (25 selected entries, branch == main)
aieweb      q1: IDENTICAL  (25 selected entries, branch == main)
aieweb      q2: IDENTICAL  (25 selected entries, branch == main)
m365dotnet  q1: IDENTICAL  (25 selected entries, branch == main)
m365dotnet  q2: IDENTICAL  (25 selected entries, branch == main)
```

q1 = "How does authentication work in this codebase?" (concept-only),
q2 = "Fix the retry logic when a request is throttled." (mixed), both with `--no-git` so
the recency signal cannot vary between arms. Byte-identical ranked selection lists on all
six pairs.

**Stated plainly so nobody reads more into it than is there:** this is a
selection-identity check on two prompts per repo, **not** a re-run of the 50-case
`tools/rank_mine_eval.py` MRR harness. That harness shells out to `neo --dry-run` once
per case; at ~12 s per invocation on this machine (load average 22–37 from concurrent
goalpool sessions) a single 50-case arm on `neo` exceeded 10 minutes, and six arms across
the three flagships was not achievable in this run. The identity check is the stronger
claim per unit of evidence — an identical selection list makes the MRR identical by
construction — but it samples two queries per repo rather than fifty. If the shepherd
wants the full M1 tables re-measured, say so and I will run them on a quiet machine.

## Review round: a fresh non-author pass, and what it found

A fresh non-author Claude verified the branch against both issues before the shepherd
sees it. It found the #196 fix had introduced **the other half of its own defect**, plus
three quieter leaks. All five are fixed in `be2427b`; the two `--- BEFORE / AFTER ---`
blocks above still hold.

**The one that mattered: on the languages this PR added, a genuine missing handler
passed silently.** `_strip_comments_and_strings` used Python's `tokenize`, which does
*not* fail on C# — `//` lexes as floor-division — so line comments survived as code and
a file carrying `// TODO: use HashSet<int> here one day` matched the uniqueness marker
while deduplicating nothing. C# `///` and TS `/** */` doc comments name `HashSet`,
`Distinct` and `OrderBy` constantly, so this was the common case. The PR had converted
"always warns, uselessly" into "silently passes whenever a doc comment mentions the
API", which is the worse failure of the two. Stripping is now per language.

**The matcher was also too weak to be useful.** A substring test over whitespace-stripped
text let `set(` match `offset(`, `reset(` and `dataset(`, and `Set<` match `Subset<` —
UNIQUE_ELEMENTS was close to inert on real Python and TypeScript, each collision
silently dropping a real warning. Markers are now anchored on the left identifier
boundary. Two things had to change together for that to be correct: normalization keeps
the single space between two identifiers (dropping *all* whitespace fuses `new HashSet<`
into `newhashset<`, so the anchor would reject C#'s primary idiom), and the right edge
stays deliberately **unanchored**, because markers are family prefixes — `OrderBy` is
written to cover `OrderByDescending`, `bisect` to cover `bisect_left`, and anchoring
both edges fires a warning at code that satisfies the constraint, which is #196
re-entering through the matcher. `frozenset(` and `new Set(` move into the tables in
their own right.

**Three honest-degradation leaks, one of which dropped a true caution.** The episode
ledger recomputed check status inline as `warning if diagnostics`, and
`aggregate_verification_status` ranks `warning` above `skipped` — so an info-only "not
checked for `<language>`" note became the run's whole `verification_verdict`, which is
#196's false alarm arriving through the ledger instead of the console. Separately, the
phase summary said "All 1 checker(s) clean", VERIFY's verdict said "nothing pushed
back", and the caution reading "No checkers on this machine. This code is unverified."
was **suppressed** — because the list was not empty. That last one is the worst shape:
a true caution dropped, not a false one added. All four now read
`_static_check_status` / the new `_checks_that_evaluated`.

Also `information` is pyright's own severity spelling, so the one real tool that emits
informational diagnostics was the one the new non-actionable rule missed.

The verifier additionally confirmed **(c)**: every read of `NeoOutput.confidence` across
`src/`, both host adapters and the CAR tool schema is either `None`-guarded or
JSON-serializes cleanly — no crash path from the `Optional[float]` change. And **(e)**:
nothing under `context_gatherer`, `index/` or the walker is touched, which Goals 3 and 5
own.

### Host adapters, which the first round missed

`--json` can now emit `"confidence": null` plus a new `"confidence_basis"`, and neither
integration surface documented it — a violation of this repo's own rule that both
adapters change when the contract does. An LLM host handed `null` under a standing
instruction to "report Neo's confidence as the number it is" has no guidance, and the
failure mode is precisely #199's: treating a run with no number as less trustworthy than
an empty patch self-reporting `0.96`. Both `.claude-plugin/agents/neo.md` and
`plugins/neo/skills/neo/SKILL.md` now name the null case and both basis values, and
`test_both_entry_points_teach_the_null_confidence_contract` pins it against the
constants so it cannot drift again.

### Verification of the review round

- [x] `tests/test_constraint_marker_matching.py` — 20 new tests: comments and doc
      comments per language, the string-literal and URL-inside-a-string cases, all five
      boundary collisions, both TypeScript `Set` spellings, the family-prefix markers
      that must stay unanchored, and one test asserting **every marker in every table
      can match its own text** — a marker no code can contain is #196 by construction.
- [x] `tests/test_static_check_honesty.py` — 5 new tests, including the episode-ledger
      verdict in both directions.
- [x] `tests/test_host_adapter_parity.py` — the null-confidence parity test.
- [x] Re-run after the fixes: the two original modules (41) + the two new ones (25) +
      host-adapter parity (54) = **120 passed**; Guard-invariant battery **48 passed,
      1 skipped**; engine/static-analysis/orchestrator-events/CAR-schema/learning-
      episodes/schemas/subcommands = **150 passed**; `ruff check src/ tests/ tools/`
      clean.

## Checklist

- [ ] I have made corresponding changes to the documentation
- [x] New and existing unit tests pass locally with my changes (one pre-existing
      environmental failure, evidenced above)

## Additional Notes

Three judgement calls worth a reviewer's attention:

1. **`NO_VERIFIABLE_CHANGE` is the floor of the trust ladder, and that is the only
   relation any code depends on.** `ANALYSIS_ONLY` is placed above
   `PROCEED_WITH_CAUTION` because it carries no unverified patch to act on — not
   because it is more certain; there is no certainty claim in it at all. An
   unrecognized action ranks below every known one.
2. **An info-only static check reports `skipped`, not `passed`.** That is a deliberate
   behaviour change beyond the reported bug: it means a constraint check that could not
   evaluate the target language no longer contributes to `static_ran_clean`, so a run
   whose only checker was unable to check anything will not early-exit.
3. **`UNSCORED_MEMORY_PRIOR = 0.5`** in `engine.py` keeps the legacy memory-entry prior
   at its old value on purpose. It is a starting weight on the retrieval layer's scale,
   not the operator-facing verdict — naming it is what keeps the two apart now that the
   run-level sentinel is gone.

### DIFF
--- a/.claude-plugin/agents/neo.md
+++ b/.claude-plugin/agents/neo.md
@@ -100,6 +100,15 @@ Then:
    plainly: "Neo found…" versus "Looking at this myself…". Never present your
    own reasoning as Neo's, or the reverse.
 6. Report Neo's confidence as the number it is. Do not round it up in prose.
+7. **`confidence` may be `null`, and that is an answer, not a gap.** It means
+   the run produced nothing a confidence number could describe — read
+   `confidence_basis` and say which: `analysis_only` (Neo answered and proposed
+   no change) or `no_verifiable_change` (Neo named files but shipped no diff,
+   so those suggestions were excluded from the score). Never substitute a
+   number, never call a `null` run low-confidence, and never rank it below a
+   scored run on the strength of the missing number — an empty patch
+   self-reporting `0.96` is the exact inversion this contract exists to stop.
+   `orchestrator.summary` already states the case in Neo's voice; lead with it.
 
 ### Voice
 
@@ -161,7 +170,7 @@ Events on stderr are `{"type": ..., "phase": ..., "message": ..., "data": {...}}
 | `hypothesis_rejected` | An approach was discarded. Worth surfacing. |
 | `risk_found` | A simulation issue or failed static check. Worth surfacing. |
 | `personality_beat` | Same line as `orchestrator.personality`. |
-| `completed` | Run finished. `data.confidence`, `data.elapsed_seconds`. |
+| `completed` | Run finished. `data.confidence` (may be `null`), `data.elapsed_seconds`. |
 | `failed` | Run raised. `data.error_type`. **No final JSON result follows.** |
 
 Exactly one of `completed` or `failed` terminates every run. If you see
@@ -203,6 +212,7 @@ Use the Neo agent to debug this race condition in the task processor.
 ## Important Notes
 
 - Neo queries take 5-30 seconds (uses LLM API calls)
-- Always verify low-confidence suggestions (<0.7)
+- Always verify low-confidence suggestions (<0.7). A `null` confidence is not a
+  low one — check `confidence_basis` before saying anything about certainty.
 - Provide rich context for better results
 - Ordinary analysis is explicit `advise` and does not mutate durable memory. Use `learn` only when the user intends to record evidence; never request `agent` without explicit workspace authority and a host executor.
--- a/plugins/neo/skills/neo/SKILL.md
+++ b/plugins/neo/skills/neo/SKILL.md
@@ -62,6 +62,16 @@ purpose:
 | `cautions` | **Never drop these.** Low confidence, failed checks, open questions. |
 | `recommended_narration` | Advisory progress lines. Reword freely. |
 
+`confidence` may be `null`, which is an answer rather than a gap: the run
+produced nothing a confidence number could describe. Read `confidence_basis`
+and report which case it is — `analysis_only` (Neo answered and proposed no
+change) or `no_verifiable_change` (Neo named files but shipped no diff, so
+those suggestions were excluded from the score). Do not substitute a number,
+do not call a `null` run low-confidence, and do not rank it beneath a scored
+run: an empty patch self-reporting `0.96` outranking a correct analysis is the
+inversion this contract exists to stop. Neo's result is an input to your work,
+not the deliverable — so state which case it is rather than papering over it.
+
 Useful event types on stderr: `memory_found` (prior facts recalled),
 `hypothesis_formed`, `hypothesis_rejected` (an approach was discarded — usually
 worth surfacing), `risk_found`, and exactly one of `completed` or `failed`
@@ -114,4 +124,5 @@ is no fallback line.
 
 - This skill uses explicit `advise` mode: it may retrieve established memory but does not create candidates or update durable memory. Use `neo --mode learn` only when the user intends to contribute outcome evidence.
 - For code review, optimization, architectural decisions, debugging, or pattern extraction, prefer the more specific Neo skills (`$neo-review`, `$neo-optimize`, `$neo-architect`, `$neo-debug`, `$neo-pattern`).
-- Always verify low-confidence suggestions (< 0.7) before applying them.
+- Always verify low-confidence suggestions (< 0.7) before applying them. A
+  `null` confidence is not a low one — read `confidence_basis` first.
--- a/src/neo/car_tool_schema.py
+++ b/src/neo/car_tool_schema.py
@@ -432,6 +432,7 @@ def neo_output_to_dict(output: NeoOutput) -> dict[str, Any]:
         "static_checks": [_static_check_to_dict(c) for c in output.static_checks],
         "next_questions": list(output.next_questions),
         "confidence": output.confidence,
+        "confidence_basis": output.confidence_basis,
         "notes": output.notes,
         "metadata": dict(output.metadata),
         "goal_assessment": (
--- a/src/neo/cli.py
+++ b/src/neo/cli.py
@@ -1420,6 +1420,7 @@ def main():
             ],
             "next_questions": output.next_questions,
             "confidence": output.confidence,
+            "confidence_basis": output.confidence_basis,
             "notes": output.notes,
             "metadata": output.metadata,
             "goal_assessment": (
@@ -1443,7 +1444,8 @@ def main():
             output.confidence,
             output.next_questions,
             output.plan,
-            output.code_suggestions
+            output.code_suggestions,
+            output.confidence_basis,
         )
         output_dict["confidence_interpretation"] = confidence_interpretation
 
@@ -1454,7 +1456,15 @@ def main():
         else:
             # Human-readable text mode
             print("\n" + "="*80)
-            print(f"CONFIDENCE: {output.confidence:.2f}")
+            if output.confidence is None:
+                # Printing "0.50" here is what made a correct analysis look
+                # like a hedged one. The interpretation carries the reason.
+                print(
+                    f"CONFIDENCE: n/a "
+                    f"({confidence_interpretation['message']})"
+                )
+            else:
+                print(f"CONFIDENCE: {output.confidence:.2f}")
             print("="*80)
 
             # The same summary a host would relay, so the terminal reader and
--- a/src/neo/config/beats/neo_matrix.yaml
+++ b/src/neo/config/beats/neo_matrix.yaml
@@ -111,6 +111,7 @@ orchestrator_voice:
     phase_checks_clean: "All {count} checker(s) clean."
     phase_checks_failed: "{failed} of {count} checker(s) pushed back."
     phase_checks_none: "No checkers here. Nothing verified this."
+    phase_checks_inconclusive: "The checkers ran and could evaluate none of it. Treat this as unverified."
     phase_checks_considering: "Considering the checkers."
     phase_checks_skipped: "Out of time before I could run them. This is unchecked."
 
--- a/src/neo/config/beats/neo_matrix.yaml
+++ b/src/neo/config/beats/neo_matrix.yaml
@@ -72,12 +72,23 @@ orchestrator_voice:
     verify_clean: "nothing pushed back"
     verify_none: "no checkers could run"
 
+    # Said instead of a number when there is no code change to score. Stage
+    # -neutral like the cautions: "I have no score for this" is a fact about
+    # the run, and a terse stage must not shorten it into looking like one.
+    confidence_analysis_only: "No confidence score — I proposed no code change."
+    confidence_no_verifiable_change: >-
+      No confidence score — I named a file and produced no change to go
+      with it.
+
     # Cautions — stage-neutral by design (see the note above).
     caution_low_confidence: "I'm at {confidence} on this. Don't take it on trust — check it."
     caution_check_failed: "{tool} isn't happy: {summary}"
     caution_sim_issues: "I ran this through and hit {count} problem(s) with my own approach."
     caution_unverified_skipped: "I ran out of time before the checkers. This code is unverified."
     caution_unverified_no_tools: "No checkers on this machine. This code is unverified."
+    caution_no_verifiable_change: >-
+      {count} suggestion(s) named a file but carried no diff and no code.
+      Nothing here is verifiable — treat this as no answer, not a safe one.
     caution_open_question: "Still open: {question}"
     caution_validation_blocked: "Proof still missing for required gate {gate_id}."
     caution_hypothesis_provisional: "Causal claim {hypothesis_id} is still {status}; do not publish it as fact."
--- a/src/neo/constraint_verification.py
+++ b/src/neo/constraint_verification.py
@@ -100,7 +100,10 @@ _PYTHON_MARKERS: Dict[ConstraintType, tuple] = {
     ConstraintType.SORTED: ("sorted(", ".sort(", "heappush", "heappop", "bisect"),
     ConstraintType.INCREASING: ("sorted(", ".sort(", "bisect"),
     ConstraintType.DECREASING: ("sorted(", ".sort(", "reverse=True"),
-    ConstraintType.UNIQUE_ELEMENTS: ("set(", "dict.fromkeys"),
+    # `frozenset(` is listed in its own right: markers are matched on left
+    # identifier boundaries, so it is no longer reached as a substring of
+    # `set(` — which is the same rule that stops `offset(` reaching it.
+    ConstraintType.UNIQUE_ELEMENTS: ("set(", "frozenset(", "dict.fromkeys"),
     ConstraintType.NON_NEGATIVE: ("abs(", "max(0"),
     ConstraintType.DIVISIBILITY: ("%",),
 }
@@ -126,7 +129,9 @@ _TYPESCRIPT_MARKERS: Dict[ConstraintType, tuple] = {
     ConstraintType.SORTED: (".sort(", "sortBy(", "orderBy(", "toSorted("),
     ConstraintType.INCREASING: (".sort(", "sortBy(", "toSorted("),
     ConstraintType.DECREASING: (".sort(", ".reverse(", "toReversed(", "orderBy("),
-    ConstraintType.UNIQUE_ELEMENTS: ("new Set", "uniq(", "uniqBy(", "Set<"),
+    # `new Set(` carries its trailing paren so it cannot be found inside
+    # `new Settings()`; `Set<` covers the annotated form.
+    ConstraintType.UNIQUE_ELEMENTS: ("new Set(", "uniq(", "uniqBy(", "Set<"),
     ConstraintType.NON_NEGATIVE: ("Math.abs(", "Math.max(0"),
     ConstraintType.DIVISIBILITY: ("%",),
 }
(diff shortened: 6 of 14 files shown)


kkeeling:
**Shepherd gate — evidence verified, flipping ready.** CI green (matrix + invariant battery). Before/after demonstration in the body covers both DoD scenarios (#199 inversion corrected on the declared trust ladder; #196 C# false-alarm gone, Python real-warning kept — plus the tokenizer-spacing defect found and fixed). The brief's M1 paste is waived by construction: the diff touches no selection surface (no context_gatherer, file_retrieval, or index/ files — verified from the file list), so retrieval metrics cannot move. Daemon review is the independent gate from here.

PR #207, opened 2026-08-12
fix(selection): --include guarantees the file, and the count says what it counts (#197, #198)

### LINKED ISSUE #197: 'Gathered N files' counts chunks, so a two-file run reports four files
`src/neo/cli.py:1098`:

```python
total_bytes = sum(gf.bytes for gf in gathered)
progress.note(f"Gathered {len(gathered)} files ({total_bytes:,} bytes)")
```

`gathered` is **chunk-level**, not file-level. Each entry carries `start`/`end` (the very next lines build `ContextFile(..., line_range=(gf.start, gf.end))`), and `MAX_CHUNKS_PER_FILE = 2` in `context_gatherer.py:27` means one file can contribute two entries. So `len(gathered)` overcounts files by up to 2x, and the label says "files".

## Observed

2026-08-11, a run asked for four specific files. The progress line read:

```
Gathered 4 files (28,975 bytes)
```

Actual content sent: **4 scored chunks from 2 files** —

```
src/.../DorothyCandidateDeriver.cs     (lines 164-360)
src/.../DorothyCandidateDeriver.cs     (lines 374-485)
src/.../DorothyInductionJobService.cs  (lines 796-850)
src/.../DorothyInductionJobService.cs  (lines 606-795)
```

Two of the four requested files (`CandidateSchemaFragment.cs`, `GapDefinition.cs`) were never sent at all, and the specific line under discussion — line 447 of a ~1000-line file — fell outside every window.

## Why it matters

"Gathered 4 files" after requesting 4 files reads as complete coverage. It is the operator's primary signal for what Neo actually saw, and here it agreed with the request by coincidence of arithmetic while the truth was half the files, partially.

This directly caused a downstream misdiagnosis: an operator concluded Neo had full context and was declining to answer for some other reason, when Neo was correctly reporting that its context was truncated. Neo's own behaviour was right; the progress line contradicted it.

This is adjacent to #187, which restores `--dry-run` to assemble the full context, and to the #178 truncation markers it references ("the markers whose entire purpose is that a cut be visible"). Same goal, different surface: #187 fixes the detailed view, this is the one-line summary most runs actually show.

## When the fix lands

A run that sends two chunks from each of two files reports two files, and says something about the chunking rather than hiding it. Suggested:

```python
file_count = len({gf.path for gf in gathered})
progress.note(f"Gathered {file_count} files, {len(gathered)} chunks ({total_bytes:,} bytes)")
```

A test asserts that gathering two chunks from one file reports `1 file`, not `2 files`.

Worth considering alongside: when a requested/allowlisted file is not sent at all, say so. Absence is currently unrepresentable in this output — the operator can only infer it by diffing what they asked for against what is listed, which the summary line does not show.

## Scope

No open PR touches this line (checked #186, #187, #190, #191, #192, #194). Confirmed present on `origin/main` at `cli.py:1098`.

### LINKED ISSUE #198: --include only filters candidates, so an explicitly named file can still be dropped or truncated silently
`--include` is documented as "Allowlist glob patterns (repeatable)" (`cli.py:470`), and is applied once, as a **candidate filter**, in `context_gatherer.py:130`:

```python
if includes and not any(fnmatch.fnmatch(rel_path, g) for g in includes):
    continue
```

After that the file is treated like any other candidate. It still faces:

- `MIN_SCORE_THRESHOLD` eviction — `scored_filtered = [... if sc >= MIN_SCORE_THRESHOLD]` (`context_gatherer.py:901`), which drops it from context **entirely**
- the `--max-files` cap (default 30 for gathering)
- the `--max-bytes` budget (default 300,000)
- `MAX_CHUNKS_PER_FILE = 2` and `MAX_MERGED_WINDOW_LINES = 200`, capping any file at ~400 lines regardless of its size

So `--include 'Foo.cs'` means "Foo.cs is eligible for selection", not "Foo.cs will be in the context". For a file over ~400 lines it is guaranteed to arrive partial, and if it scores below threshold it does not arrive at all — in both cases with no signal distinguishing that from "you didn't ask for it".

## Observed

2026-08-11. A run passed explicit includes for four files plus a raised `--max-bytes` (400,000). Two of the four never appeared in the context, and the two that did arrived as two windows each. Raising `--max-bytes` and `--max-files` changed nothing, because neither was the binding constraint — the per-file chunk cap was.

The working construction turned out to be: read the files outside Neo, inline them complete into the prompt body, and pass `--no-scan --max-files 0` to disable gathering altogether. That is a reasonable escape hatch, but it means the flag that exists to say "use these files" cannot express it.

## Two separable asks

1. **Make the semantics explicit.** If `--include` is intended as a filter, the help text should say the file remains subject to scoring, eviction, and chunking. This alone would have saved the debugging above.
2. **Provide a way to say "this file, whole."** Whether that is `--require`, or `--include` bypassing `MIN_SCORE_THRESHOLD` and the chunk cap for explicitly-named paths, is a design call. The distinction that matters is between *narrowing the search* and *asserting the inputs*.

Note the chunk caps themselves are not the complaint — the rationale in the comments ("so one large file doesn't dominate the budget") is sound. The issue is that an explicit user instruction is silently subject to them.

## Related

#194 is adjacent: it improves selection by content rather than filename/bytes, and its body observes that current heuristics "are also blind to eviction: a file pushed below `MIN_SCORE_THRESHOLD`". That is the same eviction path, approached as a scoring-quality problem. This issue is about explicit user intent surviving it, which better scoring does not by itself guarantee.

Reporting overlap: when an included file is evicted or truncated, the summary line gives no indication — see the companion issue on `Gathered N files` counting chunks.

## When the fix lands

Either the `--include` help text states the file may still be dropped or truncated, or a flag exists that guarantees a named file is sent whole and errors when the budget cannot accommodate it. A test asserts that a named file scoring below `MIN_SCORE_THRESHOLD` is either present or reported as omitted — not silently absent.

kkeeling:
## Description

Two defects in what Neo tells you it saw, both wrong in the reassuring direction.

**#197 — `Gathered N files` counted chunks.** `gathered` is chunk-level: a file over 15KB contributes up to `MAX_CHUNKS_PER_FILE` entries, so the number overstated coverage by up to 2×, under a label saying "files". On the run in the issue it read `Gathered 4 files` for four windows of **two** files, on a request that had named four — it agreed with the request by arithmetic coincidence while two named files never arrived, and the operator concluded Neo had full context and was declining for some other reason. Both numbers now, always, even when they agree:

```
[Neo] Gathered 4 chunks from 2 files (28,975 bytes)
```

The same overcount lived on two more surfaces and is fixed on both: the `--dry-run` rank listing header (`files selected, in rank order` → `12 context entries from 10 files, in rank order`) and — the one that matters most — `REPOSITORY CONTEXT (N files, …)` in the reasoning prompt, where two windows of one file were announced **to the model** as two files, inside the prompt, where no operator ever sees it to doubt it.

**#198 — `--include` only filtered candidates.** It narrowed the walk and then let the named file compete like any other, so it could be evicted below `MIN_SCORE_THRESHOLD`, dropped by `--max-files`, or capped at ~400 lines by `MAX_CHUNKS_PER_FILE` — silently in every case, indistinguishable from never having asked. Standing ruling 1 of `docs/unified-store-plan.md` makes it an assertion instead: **the named files are guaranteed whole, AND the ordinary scan still runs.**

- the walk is no longer narrowed by `--include`; matches resolve against the scanned candidates and are pinned separately (`resolve_includes`)
- pins are delivered first and whole, bypassing the score threshold, the rank ordering, the file cap and the chunk cap (`pin_included_files`)
- a **literally-named** file that the walker skipped for being over its 512 KB per-file limit is rescued with one `stat`, because that limit is a scan-cost guard rather than a relevance judgement — the same class of rule as the others a pin already bypasses. Without it the guarantee had an unstated cliff 63 KB above #198's own 449 KB case. `should_ignore` and `load_gitignore_patterns` are called rather than reimplemented, ancestor directories are tested separately, and paths outside the root are refused, so G1-inv still holds. A glob gets no rescue — expanding one means walking — and the warning names that limit
- the only thing that may still cut a pin is `--max-bytes`. That cut is marked inline with `text_budget`'s marker and reported. When several pins overflow the ceiling the budget is apportioned **max-min fair** rather than spent first-come — greedy order gives the last-named file zero bytes and no marker, which is the silent drop this issue is about, reintroduced inside its own fix
- `--max-bytes` bounds pinned **content**, and where it cannot bound the block the code says so instead of pretending. Below roughly one truncation marker per named file the block runs over the ceiling; dropping named files to make the number come out is the one move ruling 1 forbids, so `PinReport.overshoot` reports the real total and the remedy
- the prompt renderer does not apply its per-file character cap to a pin. This matters: 3,000 characters is ~40–75 lines, **less** than the windowing being complained about, so routing a pin through the ordinary cap would have delivered a "guaranteed" file more truncated than before. The 20-file display cap cannot drop a pin either
- the semantic lane honoured `--include` not at all; it pins too — a guarantee a second flag switches off is not one
- an `--include` pattern matching nothing warns, as a prompt-named path that matches nothing already did

`--include`'s help text states the guarantee instead of "Allowlist glob patterns".

## Type of Change

## Related Issue

Closes #197
Closes #198

Plan of record: `docs/unified-store-plan.md`, **Goal 3. Selection truthfulness**.

## Motivation and Context

Both issues came out of the 2026-08-11 forensic review of Neo file selection. They share a failure shape: the operator's picture of the context was wrong in the direction that reads as complete, so a correct refusal by the model looked like an unhelpful model. #197 is the one-line summary most runs show; #198 is the flag that exists to say "use these files" and could not express it.

## How Has This Been Tested?

- [x] **Guard-invariant battery** — `pytest -m invariants`: **97 passed, 1 skipped** at this HEAD (60 of them this branch's own `test_selection_invariants`; the rest arrived with Goal 5). Includes a new `TestIncludedFileGuarantee` class asserting G2-inv for `--include` per language (C#, TypeScript, Python): the named file is present, is one entry not windows, is byte-identical to disk, the scan still contributes alongside it, and **the prompt renderer sends it uncut** — the end of the pipeline, which is where the guarantee is actually spent.
- [x] **`tests/test_include_guarantee.py` — 64 new tests.** Every guarantee assertion is paired with a control that proves the unpinned behaviour still bites: `test_the_control_run_really_does_drop_it`, `test_an_unpinned_file_of_the_same_size_still_is`, `test_the_walker_really_does_skip_it`, and the chunk-cap control. A guarantee test that passes because the file would have been selected anyway asserts nothing.
- [x] **CI green** on 3.10 / 3.11 / 3.12 / 3.13 plus the Guard-invariant battery, on the head commit.
- [x] Locally, the selection-touching modules — `test_include_guarantee`, `test_selection_invariants`, `test_context_file_truncation`, `test_context_gatherer_scoring`, `test_gatherer_exclusions`, `test_file_retrieval`, `test_index_file_selection`, plus `test_dry_run`, `test_orchestrator_events`, Goal 5's `test_eligibility_differential` / `test_eligibility_single_source` and Goal 6's `test_content_index`: **405 passed** across the selection and eligibility modules.

**Test Configuration**: Python 3.13.7 · macOS 26.5.2 (Apple M4 Max) · neo 0.45.0 · merged with `origin/main` = `ebd29e0`

### Four local failures that are not this branch's

A full local run showed four `subprocess.TimeoutExpired` failures on tests that spawn `neo --json`. All four were **reproduced identically on unmodified `origin/main` (4a48b80)**, in a separate worktree with its own venv — they are an environment property of this machine, and CI passes all four on every Python version.

| Test | this branch | `origin/main` @ 4a48b80 |
|---|---|---|
| `test_cli_global_flags.py::…::test_no_scan_dry_run_does_not_require_api_key` | timeout | timeout |
| `test_integration.py::…::test_consolidation_performance_with_all_boosts` | timeout | timeout |
| `test_learning_evaluation.py::test_evaluation_cli_runs_without_provider_keys_or_user_memory` | timeout | timeout |
| `test_operating_modes.py::test_verify_cli_needs_no_provider_and_agent_cli_fails_before_provider` | timeout | timeout |
| `test_dry_run.py::…::test_dry_run_prints_the_assembled_prompt_not_just_files` | **fixed here** | passed |

The last row was a real break and is fixed rather than waived: that test asserted the literal header `files selected, in rank order`, which is the #197 mislabel. The assertion now pins the truthful form.

## M1 — retrieval quality, branch vs main

Instrument: `tools/rank_mine_eval.py`, per `docs/eval-baselines-2026-08.md`. Both arms of a repo share `--repo` (identical mined cases at an identical HEAD) and differ only by `--tree`, so the delta is the ranker and nothing else. 50 cases per arm (harness default), `--skip-recent 50`, git recency off (the tool's default; `--with-git` opts in), `--k 1 3 10`, `--timeout 600`.

Re-measured in full against `ebd29e0`, so both arms include Goal 5's walker refactor (#208) and Goal 6's persistent content index (#209), and the table describes the code under review. Every arm has been re-run after each review round and each merge; this is the reading at the head commit `67542a7`.

| repo | arm | n | failed | **MRR** | R@1 | R@3 | **R@10** | H@10 |
|---|---|---|---|---|---|---|---|---|
| neo | main `ebd29e0` | 50 | 0 | 0.709 | 0.311 | 0.504 | 0.681 | 0.960 |
| neo | **branch** `67542a7` | 50 | 0 | **0.709** | 0.311 | 0.504 | **0.681** | 0.960 |
| aieweb | main `ebd29e0` | 50 | 0 | 0.728 | 0.487 | 0.654 | 0.778 | 0.880 |
| aieweb | **branch** `67542a7` | 50 | 0 | **0.728** | 0.487 | 0.654 | **0.778** | 0.880 |
| m365dotnet | main `ebd29e0` | 50 | 0 | 0.669 | 0.312 | 0.497 | 0.771 | 0.860 |
| m365dotnet | **branch** `67542a7` | 50 | 0 | **0.669** | 0.312 | 0.497 | **0.771** | 0.860 |

**Identical on every metric, on all three flagships.** No regression, and no improvement — this goal should not move ranking, and it does not.

Repos measured at: neo `ebd29e0`, aieweb `26fff07e0f4a`, m365dotnet `61dc4a171bdf`.

**The m365dotnet main row was measured twice, and the first reading is why.** Run back-to-back with the branch arm it reads 0.669 / R@1 0.312, exactly as tabled. Run ~35 minutes earlier it read **0.659 / R@1 0.292** — one case of fifty, ranked 2nd instead of 1st. The harness is not fully deterministic across time: `_history_boost` reads the FactStore, `rank_mine_eval.py`'s own docstring notes it stays live under `--no-git`, and the observer daemon writes to `~/.neo` on its own schedule. So the ambient drift on this repo is about ±0.01 MRR, or one case. Rather than quote the flattering arm — the earlier reading made the *branch* look better — the main arm was re-run adjacent to the branch arm and both agree exactly. A later goal comparing against this table should read a one-case move as noise, and should measure its two arms close together.

That drift is also why the selection-identity check below carries the weight here: it is exact, and it does not depend on the FactStore at all.

Two more things to record so a later goal can reproduce this rather than trust it:

- **The neo arm uses a clean detached worktree at `origin/main`, not the platform-root `neo` checkout**, which was dirty and under active edit by another agent while these ran. `--repo` is the same path for both neo arms, so the case set is identical.
- These absolutes are **not** comparable to `docs/eval-baselines-2026-08.md`'s M1 table. That document was produced by `tools/rank_eval.py` at `5bbee46` (in-process `score_candidate`); this is `rank_mine_eval.py` (shells out to the real CLI), which the baselines doc itself warns is a different instrument reporting different absolutes. The A/B above is internally consistent, which is what a no-regression claim needs.

### Selection identity — the structural proof behind the numbers

Metrics agreeing to three decimals is strong; identical output is stronger, and here it is also the more robust instrument. The canonical M2 battery prompts, run against m365dotnet under both trees, comparing the full `--dry-run` selected-file listing (path, line range, bytes and score, in rank order):

| prompt | shape | entries | verdict |
|---|---|---|---|
| P1 | file-named | 30 | byte-identical |
| P2 | file-named | 30 | byte-identical |
| P3 | concept-only | 30 | byte-identical |
| P4 | concept-only | 30 | byte-identical |
| P5 | mixed | 30 | byte-identical |
| P6 | symptom | 30 | byte-identical |

Re-run after every review round and after the merge, at the head commit, with the same result each time. Which is what the code says should happen: with no `--include` on the command line, `config.includes` is empty, the walk receives exactly the arguments it received before, `resolve_includes` returns nothing and the pin list is empty. Selection for an unpinned run is unchanged **by construction**, and this is the measurement that says so rather than the argument.

## G2-inv — 100% named-file presence

**In the battery** (`pytest -m invariants`): `TestIncludedFileGuarantee`, four assertions × three languages, all green — the named file is present, is one entry rather than windows, is byte-identical to disk, the scan still contributes alongside it, and the prompt renderer sends it uncut.

**Live on m365dotnet**, four `--include` files against a prompt about none of them (one of them 449 KB — the shape of the run in #198):

| named file | main `4a48b80` | branch `6c7a3c7` |
|---|---|---|
| `…/Scheduling/SchedulingExecutionService.cs` (449 KB) | present as **one 106-line window** (7459–7565) | present, 296,808 B, **cut by `--max-bytes` and marked** |
| `src/Ink.Core/Models/AuditEntry.cs` | **ABSENT, silently** | present, whole (627 B) |
| `src/Ink.Core/Models/ChatMessage.cs` | present, whole | present, whole |
| `src/Ink.Core/Models/ChatStreamEvent.cs` | present, whole | present, whole |
| **named-file presence** | **3 / 4 = 75%** | **4 / 4 = 100%** |
| summary line | `Gathered 4 files (15,376 bytes)` — for **3** distinct files | `Gathered 4 chunks from 4 files (299,959 bytes)` |

Both defects reproduce on `main` in a single run, and #197 makes #198 invisible: "Gathered 4 files" after naming four files is exactly the reading that stops you looking for the missing one.

Everything the branch run did to the operator's picture, it said out loud:

```
[Neo] --include pins 4 file(s) 3 whole, 1 cut and marked (299,959 bytes), ahead of the scan
[Neo] Warning: --max-bytes=300,000 forced a cut in 1 pinned file(s)
      (src/Parslee.M365.Api/Services/Scheduling/SchedulingExecutionService.cs) - marked inline
[Neo] Gathered 4 chunks from 4 files (299,959 bytes)
```

…the cut is marked in the prompt itself (`... [truncated: 162462 of 458333 characters not shown]`), and the banner counts what was sent rather than what was offered.

**G3-inv**: every cap that fired names itself, every cap that did *not* fire stays quiet, and the counts in the gather line, the dry-run listing and the prompt banner all count distinct files and chunks separately. Two honest limits on that claim, both found by review and both now true rather than glossed: the byte cap is detected where the loop actually rejects a candidate (not inferred from a running total that can sit below the ceiling while the ceiling turns everything away), and `--max-bytes` bounds pinned *content* — where it cannot bound the block, the overshoot is reported rather than papered over.

## Merged with Goals 5 and 6

`origin/main` gained the one-walker refactor (#208) and then the persistent BM25 content index (#209) while this was in review. Both merged in, `8d6efa9` and `fdbd0e6`.

**#208** moved the walk, the ignore rules and the gitignore matcher into `neo.eligibility`. The `--include` rescue now calls `eligibility.should_ignore` / `eligibility.load_ignore_patterns` — the same shared module the walk uses, so it still holds no eligibility logic of its own and G4-inv's "exactly one implementation" survives. `WalkPolicy.skip_symlinks` defaults to True, so the shared walk refuses symlinks where `os.walk` listed them; the rescue refuses them too, and `test_the_rescue_matches_the_walker_on_symlinks` followed the walk across that change on its own, which is what it was written for. **#209** split the walk into `base_paths` (the repository's corpus, no per-call flags) and `filter_candidates` (this call's flags applied to it) — a better home for a decision this branch had been reconstructing. `--include` overrides `--exts`, and until then the only way to honour that was to let the ext-filtered candidate list miss the file and re-`stat` it in the rescue. The pin pool is now `filter_candidates(eligible, [], excludes, None)`: exclude- and gitignore-filtered, ext-unfiltered, by construction. The rescue narrows to what it is actually for — a file over the walker's 512 KB ceiling, which is where #198's own 449 KB case sits just under. Its two ceiling tests moved to that path, because they had been reaching it via `--exts` and were passing on the wrong mechanism.

Goal 5's differential and single-source gates and Goal 6's `test_content_index` all pass against this branch.

## Review rounds — fourteen defects found and fixed

Two fresh non-author reviews found fourteen defects between them, every one in code this branch added and nearly all of them the species this goal exists to delete — a report claiming more than happened, or a cap blamed for an absence it did not cause. Each is fixed with a test, and each was verified against the reviewer's own reproduction.

**Round one** (fixed in `6c7a3c7`):

1. **`--max-bytes` documented as a hard ceiling and exceeded 2×.** 200 pins against `--max-bytes 5000` delivered 10,200 bytes of pure markers and no source. Now reported with the real total and the remedy, and the docstring says what the number actually bounds.
2. **The starved-scan warning named caps that had not bound.** A three-file repo with 135 bytes pinned against a 300,000-byte ceiling was told its budget was used up. It now names only the cap that bound, and where none bound it makes a bare statement with no remedy attached — this repo's own rule about never blaming a cap for an absence it did not cause.
3. **`--include` had the unstated 512 KB cliff** described above, and blamed four causes it had never checked.
4. **The pin summary said "whole" for files it had cut.**
5. **`--semantic` walked the whole tree on every run** — the `iter_paths(...)` argument was positional and so eager, even with no `--include`, on the lane whose purpose is avoiding the walk. Guarded. Pin-owned chunks are also dropped before MMR packing rather than after, so the lane stops spending budget on chunks it discards.
6. **#197's defect survived in `REPOSITORY CONTEXT (N files, …)`** — the banner the model reads.

**Round two** (fixed in `02fa070`), and the first two of these blocked on their own:

7. **The rescue minted a duplicate.** It returned the operator's raw spelling while the walk uses the normalized form, and `pinned_rels` is keyed on it — so `--include ./app.py` sent one file to the model twice, announced as two. A G1-inv duplicate created while closing a G2-inv absence, from the spelling shell completion and `find .` emit.
8. **The byte cap could bind and be denied.** The inner tests are `total + this_one > max_bytes`, so the ceiling can turn away every candidate while the total sits below it; the run then said no setting would change the outcome, one byte from `--max-bytes + 1` changing it. The old test sat exactly on the `>=` boundary, which is why it never saw this.
9. **`N files truncated` still counted chunks** — the same conflation, in the third clause of the sentence whose first clause round one had just fixed, which is what made it read as a contradiction. Its test re-implemented the buggy predicate, so it codified the defect instead of catching it.
10. **`--include` vs `--exts` was decided in code and denied in the docstring.** Resolved where ruling 1 points: `--exts` narrows the search, `--include` asserts the inputs, so an exactly-named file overrides it — documented and tested rather than incidental.
11. **The file-cap warning named a knob that could not move it** (`--max-files=<adaptive_limit>`, a number the operator never typed; below specificity 10 the adaptive limit ignores the flag entirely).
12. **An unreadable named file was a fully silent drop** — no pin note, no warning, nothing.
13. **The rescue removed an implicit size bound**, so `--include` on a multi-gigabyte artefact would read it into memory twice. `_PIN_RESCUE_MAX_BYTES` (64 MB) replaces it.
14. **A real filename containing glob punctuation** (`a[1].py`) was classified as a glob and denied the rescue. The test is now "does this name a file", which `os.path.isfile` answers exactly.

**Round three** (fixed in `67542a7`):

15. **The semantic lane's pin pool was still `--exts`-filtered**, so `--exts py --include '*.txt'` pinned on the keyword path and reported "matched no file" on the other. Both lanes build the identical pool now. That lane had no test of any kind — which is how it kept the old behaviour through the merge that changed the other one — and has three.
16. **#197's overcount was still live on the machine-readable surface.** The phase event's `file_count` was chunk-level, so a run whose dry-run header correctly read `25 context entries from 17 files` emitted `"file_count": 25` to the host — the surface both plugin adapters consume. 17 and 17 now.
17. **`--no-scan` silently voided `--include`.** Two reasonable flags combining into a named file with no signal: #198 reached from the other direction. It says so now.

## Checklist

- [x] I have made corresponding changes to the documentation (`--include` help text; battery docstring)
- [x] My changes generate no new warnings (`ruff check src/ tests/ tools/` clean)
- [x] New and existing unit tests pass locally with my changes (see the control table above)

## Additional Notes

**Scope held.** No eligibility logic is written here. Goal 5 landed mid-review and moved the walk into `neo.eligibility`; after the merge this branch calls `eligibility.should_ignore` and `eligibility.load_ignore_patterns` and restates no glob, anchoring or negation rule of its own, so G4-inv's "exactly one implementation" holds. `git diff 9b0c16d HEAD` touches seven files, all this branch's own — `eligibility.py`, `project_index.py`, `architecture_metrics.py`, `constraint_verification.py` and the rest of #206/#208 are byte-identical to main. What the branch decides is which of the *gatherer's own budget rules* an explicitly-named file may skip, which is the question #198 asks.

**Two known limits, disclosed rather than left to be discovered:**

- **The guarantee is over EXACT PATHS; a glob is a search.** An exact path — absolute or relative, the form an operator uses to name a specific file — is guaranteed, including past `--exts` and past the walker's 512 KB scan limit, and a path refused for a nameable reason (symlink, ignore rule, escape, the 64 MB ceiling) is reported with that reason rather than as "check your spelling". A glob gets neither, because expanding one means a second walk. **One case of that is genuinely silent and is worth stating plainly**: the rescue is attempted only when a pattern matched *nothing*, so `--include '*.py'` beside a 700 KB `huge.py` pins the small files and drops the large one with no signal, where `--include huge.py` pins it whole. Closing it needs the walk to report which paths its size ceiling skipped — `neo.eligibility`'s to give, and this goal is scoped out of that module. An earlier draft of this section claimed "the warning says which case you are in", which is true when a glob matches nothing and false when it matches some; the claim is corrected rather than the caveat buried.
- **`--no-scan` voids `--include`**, because the scan is what delivers it. The run now warns when both are passed, rather than producing a named file that is silently absent.
- **Under `--json` (which implies `--quiet`) none of these warnings reach the caller.** There is no event and no JSON field for them, so "never a silent drop" holds on the interactive stderr path and not on the programmatic one. This is a pre-existing pattern — the prompt-named-path warning has behaved this way since it shipped — and both plugin adapters invoke `neo --json`, so it is worth closing. It belongs to Goal 8 (one front door) rather than here, and is called out so it is a decision rather than an oversight.

One judgement call worth a shepherd's eye: `--include` is now a stronger instruction than "allowlist filter", so a broad glob like `--include '*.py'` pins every match, bypassing both `--max-files` and the renderer's display cap. That is ruling 1 applied literally, and the run reports the consequences in full (pin count, bytes, any forced cut, any starved scan) — but there is no longer a flag that means "narrow the search without asserting the inputs". If that narrowing is wanted back, it should be its own flag rather than a softening of this guarantee.

Opened as **DRAFT** per the plan's Review & merge governance (2026-08-12): the shepherd verifies the evidence and flips it ready; the car-pr-review daemon is the independent gate after that.


kkeeling:
**Shepherd gate — evidence verified, flipping ready.** CI green. M1 in the body: byte-identical on every metric across all three flagships at head, with exemplary drift disclosure (re-measured the arm whose earlier reading flattered the branch). --include guarantee + truthful counts implement the plan's standing ruling 1. Daemon review is the independent gate from here.

PR #208, opened 2026-08-12
refactor(eligibility): one walker for the gatherer, the index and the arch scan

kkeeling:
## Description

Unified-store **Goal 5 (One walker)**. The gatherer and the index builder each carried their own answer to "which paths exist and are eligible" — their own walk, their own exclusion list, their own security checks. `architecture_metrics` held a third exclusion list nobody had noticed. This extracts `src/neo/eligibility.py` as the single implementation and deletes every now-redundant copy.

`neo.eligibility` owns: the filesystem walk, `.gitignore`/`.ignore` semantics (moved verbatim from #186), the shared derived-artifact defaults, symlink rejection, and the byte-identical dedup primitive. Consumer-specific rules stay consumer-side as `WalkPolicy` knobs — the gatherer's 512 KB ceiling and `fnmatch`-dialect `--include`, the index's per-language globs — so a file missing from context can be attributed to the rule that dropped it rather than blamed on the repo's `.gitignore`.

Deleted as redundant: `ProjectIndex._build_exclusion_filter` and its glob-then-filter walk, the index's separate symlink/is_file/outside-repo checks, and `architecture_metrics._IGNORE_DIRS` plus `_safe_walk`.

## Type of Change

## Related Issue

Plan of record: `docs/unified-store-plan.md`, "Goal 5. One walker". Closes the drift that produced #159 and #186.

## Motivation and Context

The two copies drifted independently for months and broke in independent ways: #159 — the index indexed 83 Python files of a stale worktree for a repository of 4,272 C# files, and exited 0. #186 — the gatherer served gitignored build output as context. Each was fixed on its own side, and each fix left the other copy wrong; #186's PR message documents the exclusion list that had silently diverged.

Unification is what makes a third drift structurally impossible rather than merely unlikely. A reviewer cannot see a second list by reading a diff that does not contain it, so the guard has to be a test.

## M1 — retrieval quality, branch vs main

`tools/rank_mine_eval.py --cases 50`, 50 git-mined cases per repo, same harness and parameters on both arms. `--tree` pins which source tree executes, so neither arm runs the other's code.

| Repo | MRR (main) | MRR (branch) | R@10 (main) | R@10 (branch) |
|---|---|---|---|---|
| neo | 0.721 | **0.721** | 0.714 | **0.714** |
| aieweb | 0.728 | **0.728** | 0.778 | **0.778** |
| m365dotnet | 0.659 | **0.659** | 0.771 | **0.771** |

Identical on all three flagships, to three decimals, on every reported k (R@1 and R@3 match too). A pure refactor must not move selection quality, and this one does not.

## M2 — wall-clock and peak RSS, branch vs main

`tools/m2_battery.sh`, canonical 6-prompt battery on m365dotnet, 3 timed runs per prompt after a discarded warm-up, observer disabled. Arms run sequentially so they do not contend.

| | main | branch | delta |
|---|---|---|---|
| battery median wall | 62.05s | 59.44s | −4.2% |
| battery peak `ru_maxrss` | 1781.0 MiB | 1822.5 MiB | +2.3% |
| selected entries | 180 | 180 | — |
| battery union distinct files | 111 | 111 | — |

Both deltas are inside the noise floor: main's own runs ranged 43.23s–148.07s across the battery, so a 2.6s median difference is not a signal. This is a developer laptop, not an isolated box.

**The stronger result is selection invariance.** The set of files selected is byte-identical across all 18 runs and in the battery union — `diff` of the sorted 111-file unions is empty, and all 18 per-run file sets match exactly. Same paths, same order, same counts.

## G4-inv

- **One eligibility implementation** — `tests/test_eligibility_single_source.py` (AST-based, not grep, so prose cannot trip it). Guards one definition site per named function, one `DEFAULT_IGNORE_PATTERNS`, one `os.walk`, no second exclusion list detected *by content* (a copy always renames the variable), and that all three consumers still import the shared module.
- **0 over-exclusion on the differential corpus** — `tests/test_eligibility_differential.py` runs the walk against `git check-ignore` over per-test fixture repos plus this checkout. On synthetic corpora the assertion is absolute: zero tracked files skipped.

Both new test files carry `pytest.mark.invariants` and run in the ordinary suite.

**Known accepted limit, documented in the test:** nested `.gitignore` files are not read, only the repository root's. That produces *under*-exclusion (the walk yields a file git ignores), never over-exclusion, and it is the only source of under-exclusion measured against `git check-ignore` over 7,534 on-disk paths in 33 repositories. `test_nested_gitignore_is_the_documented_under_exclusion` pins it as a live fact, so closing it later is a visible change rather than a silent one.

## Behavioural deltas, stated rather than smuggled

The constraint was a pure refactor. Every delta the unification forced is listed here; none is left to be discovered.

**1. `--exts` no longer widens when handed nothing** *(a defect this branch introduced and then fixed — see `e0403ee`)*. `normalize_exts` returned `None` — the sentinel for "no extension filter" — whenever the normalized set came out empty, and dropped empty-string members. Pre-unification `exts=[]` matched **nothing**; the first cut made it match the **entire repository**, and turned `exts=["py", ""]` into `{"py"}`, discarding a real selector (`""` is the extension of `Makefile`, which is why `--exts py,` used to admit one). Now only `None` means "no filter". Reachable through the public `walk_paths`/`iter_paths` API; `cli.py` guards `--exts ""` with a falsy check, so the flag itself was never affected.

**2. `--exts` is now case- and dot-insensitive** — kept, because both spellings were plainly broken: `--exts .py` previously matched *nothing at all* (the leading dot was compared against a stripped extension), and the comparison was case-sensitive, so `a.PY` was invisible. Resolved toward correctness.

**3. The gatherer now rejects symlinks**, which the index already did and it did not. A symlink is not read, so nothing outside the repository can reach a prompt through one. It is a `WalkPolicy` flag, not an ignore rule, and the differential test states it as such.

**4. The architecture scan's eligibility widened past the old `_IGNORE_DIRS`.** It now additionally honours the repo's root `.gitignore` and `.ignore`, gains ~17 default patterns the old set lacked (`.worktrees`, `obj`, `bower_components`, `Pods`, `Carthage`, `.next`, `.nuxt`, `.svelte-kit`, `.output`, `.nox`, `.eggs`, `.idea`, `.vscode`, `.vs`, `htmlcov`, `*.egg-info`, `**/.claude/worktrees`), and rejects symlinks. Consequence: a repo that gitignores generated code now has that code dropped from god-file / complexity / coupling metrics, so those numbers can move across this commit. On this checkout the file set is unchanged (205 == 205), which is why no test caught it — hence stating it here.

**5. The index's `excluded` count changed meaning *and* scope.** The shared walk PRUNES an excluded subtree rather than globbing the whole repo and filtering afterwards, so it does not know how many files are inside and does not claim to; the report now carries `excluded_dirs` and `excluded_files` separately. Separately, `excluded_files` increments for any ignored file regardless of `match_globs`, where the old count only saw paths matching `file_patterns`. Measured on this checkout: `excluded` went **6637 → 12** (`excluded_dirs=11`, `excluded_files=1`) while the selected set stayed identical at 205 files. The operator-facing number moved by ~500×; nothing about selection did. The old "200 paths excluded" was only available because the old code walked every worktree copy in order to count what it was about to throw away.

**6. Results are sorted by `rel_path`.** Directory-entry order is whatever the filesystem hands back, so without this a tie anywhere downstream — two files at the same relevance score, two chunks at the same rank — resolves by inode order, and the same repository selects differently on two machines.

**7. Policy rejections are not counted.** A file dropped by `skip_symlinks`, the size ceiling or the extension filter is not tallied into `excluded_*`, which counts ignore-rule verdicts only. A skipped symlink is logged at DEBUG so an operator asking "why isn't this indexed?" still gets an answer — the index used to warn here, and after the move it was silent.

**Removed without a shim:** `neo.context_gatherer.load_gitignore_patterns` and `neo.context_gatherer.should_ignore` were module-level public names. All in-repo callers are updated; no deprecation alias was left, so an out-of-tree importer would break. Deliberate — leaving re-export shims for internal helpers of a CLI tool preserves exactly the ambiguity this goal exists to remove.

## How Has This Been Tested?

- [x] Full suite: **2417 passed, 29 skipped, 2 failed** — both failures pre-existing and unrelated (below)
- [x] `ruff check src tests tools` — clean
- [x] Differential vs `git check-ignore`: fixture corpora + this checkout
- [x] Structural single-source guard, AST-based
- [x] M1 on three flagship repos, both arms
- [x] M2 battery on m365dotnet, both arms, plus selected-file-set diff

**Both suite failures reproduce on the main baseline and are not from this branch:**

- `test_consolidation_performance_with_all_boosts` — a 100ms latency bound. Branch 246.3ms; **main baseline 296.9ms on the same machine**, i.e. main is *slower*. Machine speed, not a regression.
- `test_no_scan_dry_run_does_not_require_api_key` — passes in isolation, fails only in full-suite order. Pre-existing test pollution, untouched by this diff.

**Guards proven non-vacuous by mutation**, since a structural guard that cannot fail reads as coverage while asserting nothing:

- Reverting `normalize_exts` to the collapsing form → 3 failures in `TestExtensionFilterNarrows`.
- Reintroducing `os.scandir` + `Path.rglob` into `architecture_metrics` → 2 failures in `TestNoSecondWalk`.
- Pointing the single-source scan at the pre-refactor tree → all three guards fail with precise messages (`should_ignore is defined in src/neo/context_gatherer.py:303`; `os.walk … also called by ['architecture_metrics.py', 'context_gatherer.py']`).
- Regressing `compile_glob` to `fnmatch.translate` (the exact #186 defect) → `test_zero_over_exclusion[star-does-not-cross-a-separator]` fails.

**Test Configuration**:
- Python version: 3.13.7
- OS: macOS (darwin 25.5.0)
- Neo version: 0.45.0

## Checklist

## Additional Notes

Opened as **DRAFT** per the plan's 2026-08-12 governance section: the shepherd verifies the evidence above against the plan and flips it ready; the car-pr-review daemon is the independent gate after that. Not to be merged by its author.

Diff surface is confined to walking/eligibility, since Goal 3 is concurrently editing `--include`/reporting paths in the gatherer.

Findings 4, 5 and the `--exts` inversion were surfaced by a fresh non-author review pass after the implementation was complete, not by the implementer.


kkeeling:
**Shepherd gate — evidence verified, flipping ready.** CI green (matrix + battery). M1 identical to three decimals on all three flagships with byte-identical selection across 18 runs — pure-refactor claim proven, not asserted. M2 within noise. Note for the record: this PR's main-arm measurement confirms m365dotnet MRR 0.659 ≥ the 0.60 plan target — M1's headline goal is met as of #194's merge. Daemon review is the independent gate from here.

PR #209, opened 2026-08-13
perf(index): persist the BM25 content index instead of rebuilding it per call (#195)

### LINKED ISSUE #195: BM25 index is O(N) in memory and query time; ~15s and ~500MB at 5,000 files
Deferred from #194 review. Not that PR's job — the fix is in `src/neo/memory/bm25.py`, which has other callers.

## Measured

Build + one query + `tracemalloc`, on the file-selection workload:

| files | build | query | retained | tokens |
|---|---|---|---|---|
| 269 | 0.78 s | 16 ms | 27 MB | 495 K |
| 1,076 | 3.12 s | 64 ms | 108 MB | 1.98 M |
| 3,228 | 9.42 s | 195 ms | 321 MB | 5.95 M |

Clean linear: **~3 s and ~100 MB per 1,000 files.** At 5,000 files that is **~15 s and ~500 MB on every invocation**, before the LM call.

## Cause

`BM25.__init__` retains `self.docs` — every token of every file — and `scores()` calls `score()` per document, each rebuilding `Counter(doc)`. That is fine for what it was written for: a ~850-fact scope cap. It is now handed a 6-million-token corpus by `neo.file_retrieval`.

Fix is a precomputed per-doc `Counter` at build time plus an inverted index, which removes both the retention and the per-query rebuild.

## Blast radius, stated so it is not over-fixed

- `gather_context` is **CLI-only** (`cli.py`). The observer daemon does not call it, so the documented observer RSS behaviour is unaffected.
- The realistic worst case is the .NET repo CLAUDE.md already describes (4,272 C# files), not a hypothetical.
- At neo's own ~300 files this costs 0.14–0.78 s and nobody notices.

## Not urgent

`iter_paths` skips files over 512 KB, and the 200 KB read cap bounds per-file cost. This bites on file *count*, not file size.

kkeeling:
## Description

Moves #194's per-call BM25 content index onto disk, into the repository's own `.neo/`
beside the semantic catalog, with inline incremental freshness. Closes #195. Goal 6 of
`docs/unified-store-plan.md`.

#194 made file selection BM25 over file CONTENT and took m365dotnet from MRR 0.051 to
0.607. It also read and tokenized every eligible file on every invocation, which then
became the entire cost of a Neo call: a **53.47 s median wall / 1.95 GB peak RSS** to
assemble context on m365dotnet, before a single token reached the model. None of that
work is per-call — the corpus changes when the repository changes, which between two
invocations is usually not at all.

- **`index/content_index.py`** — postings, per-file content hashes, doc lengths and a
  tokenizer/schema signature in SQLite. SQLite rather than the catalog's JSON because
  the access shapes are opposite: the catalog is read whole, a keyword query touches
  ten terms out of a few hundred thousand, and a parse-whole format would spend the
  warm budget deserializing postings no query asked for. **Index artifacts only** —
  never a file body (plan ruling 2); delivery still reads whole files fresh from disk.
- **`index/freshness.py`** — one implementation of "what changed since I last looked?",
  now shared with `ProjectIndex.check_staleness`. size + mtime is the cheap pre-filter
  that decides *what to hash*; only a hash mismatch counts as a change, so `touch`
  re-stamps and re-tokenizes nothing.
- **Eligibility comes from the #208 walker and is never recomputed** — a file the
  walker stops admitting (newly gitignored, deleted) has its postings dropped on the
  same pass, so a stale index cannot answer with an excluded file.
  `test_the_module_does_not_walk_the_filesystem` fails on an `os.walk`/`glob` here.
- **One walk per invocation, unfiltered**; `--exts`/`--include`/`--exclude` narrow the
  answer afterwards. The index is a property of the repository, so `--exts py` can no
  longer prune it and force the next call to rebuild.
- **`--dry-run` reports which of cold / rebuilt / incremental (N files) / warm / memory
  happened** (G3-inv), in the selected-files block and in the `--json` payload's
  `content_index` key — `--json` implies `--quiet`, so the stderr note is suppressed on
  exactly the path a machine consumer reads.
- `EligiblePath` now carries `mtime_ns`: the walk already stats every file, and
  `os.path.getsize` was discarding the half an incremental index needs.

## Type of Change

- [x] Performance improvement

## Related Issue

Fixes #195

---

## Evidence

Full tables, exact commands and repo HEADs:
**`docs/goal6-content-index-measurements-2026-08-12.md`**.
Arms: `main` = `9b0c16d` (post-#208), `branch` = `06c3ae9`.

### M1 — retrieval quality: **identical**, all three flagships

`tools/rank_mine_eval.py`, 50 git-mined cases per repo, `--no-git`, 0 failed cases.

| repo | arm | MRR | R@1 | R@3 | R@10 | H@10 |
|---|---|---|---|---|---|---|
| neo | main | 0.712 | 0.307 | 0.502 | 0.708 | 0.980 |
| neo | **branch** | **0.712** | **0.307** | **0.502** | **0.708** | **0.980** |
| aieweb | main | 0.728 | 0.487 | 0.654 | 0.778 | 0.880 |
| aieweb | **branch** | **0.728** | **0.487** | **0.654** | **0.778** | **0.880** |
| m365dotnet | main | 0.669 | 0.312 | 0.497 | 0.771 | 0.860 |
| m365dotnet | **branch** | **0.669** | **0.312** | **0.497** | **0.771** | **0.860** |

Byte-identical in every cell. **Disclosed measurement condition:** m365dotnet is a live
tree on a shared machine and an unrelated session ran `git reset` in it at 20:58:59
(reflog), between the m365dotnet main arm and the branch arm — HEAD did not move, a
handful of untracked/modified files did. The arms agree in every cell anyway, which says
the perturbation was immaterial to these 50 cases, not that it could never matter.

Stronger still, on the M2 battery: all six canonical
prompts selected **the same files in the same rank order** on both arms (`diff` of the
per-prompt selected-file lists is empty, six for six).

**Staleness test.** Ten `.cs` files in m365dotnet given a unique marker each, index
refreshed, marker searched:
`S3 files matching the unique edit token: 1 -> ['src/Ink.Core/Editing/ApplyEditPlanResult.cs']`.
New content findable, attributed to exactly the file holding it. The converse — old
content ceasing to be findable — is pinned by
`test_an_edit_is_reflected_with_no_stale_hits`, which asserts both halves, because an
index that only ever ADDS postings passes the first check while answering with text
that is no longer in the file. All ten files restored byte-for-byte; m365dotnet's
`git status --porcelain` identical before and after.

### Post-review fix: a fresh non-author verification pass found five defects

The routing protocol requires a fresh non-author Claude to verify; it did, and it
earned its keep. All five are fixed in `260e56c`, each with a test.

1. **CRITICAL — a peer holding the write lock made Neo DELETE the index.**
   `sqlite3.OperationalError` ("database is locked") is a SUBCLASS of
   `sqlite3.DatabaseError`, and `refresh` caught `DatabaseError` first, so lock
   contention ran the corruption handler and `os.unlink`ed a perfectly good store.
   The `except sqlite3.Error` clause below it — comment: *"Locked by a peer ...
   serve this call from memory"* — was unreachable. Reproduced: the store is
   deleted and recreated, and the PEER then commits into an unlinked inode, so its
   write vanishes with no error anywhere. On m365dotnet each collision costs a
   109 MB store and a 122 s rebuild. Two Neo invocations in one repo (editor plugin
   + shell) is ordinary. `_connect` had the guard right; `refresh` did not.
   Mutation-verified: `test_a_peer_holding_the_write_lock_does_not_delete_the_store`
   fails on the original clause ordering. (The first mutation I tried swapped the
   clause headers without their bodies and passed — a vacuous check, caught and
   redone.)
2. **HIGH — every warm call opened a write transaction**, rewriting the unchanged
   signature, so two ordinary overlapping calls contended on the steady-state path.
   That is what made #1 reachable in normal use, not just during a rebuild.
3. **MEDIUM — a filtered call was re-ranked.** Corpus statistics were repo-global
   while the per-call index computed them over exactly the files that call
   admitted, and BM25 reads all of N, df and avgdl. Unflagged runs matched main
   exactly; `--exts py` changed all 25 selected lines. Now byte-identical under
   none / `--exts py` / `--exts py,md` / `--exclude docs` / `--exclude
   src/neo/memory` / `--include 'src/neo/*.py'` — table in the measurements doc.
   The "one stated difference" caveat this PR previously carried is gone, because
   the difference is gone.
4. **MEDIUM — `chmod 000` deleted a file from the corpus permanently.**
   `detect_changes` dropped any candidate it could not hash, and the caller reads
   "absent from the walk's set" as removed. The per-call index still ranked such a
   file on its path tokens. Permission was withdrawn from the CONTENT; the name is
   still a real name in the repository.
5. **LOW — the corruption flag was sticky**, so a reused `ContentIndex` reported
   `rebuilt` and wiped the repository on its second refresh.

Plus, from the same pass: the cold-build announcement now precedes the full-repo
sha256 instead of following it in silence, and term ids resolve per query rather
than preloading a few hundred thousand `terms` rows on the warm path. Fixing #1
exposed a latent bug of its own — the vocabulary was loaded BEFORE the wipe on a
rebuild, so `_index_file` believed every term already existed and wrote postings
pointing at deleted ids.

Two doc claims were false and are corrected: the measurements doc said the
concurrent path was "handled and unit-tested" (it was neither), and
`filter_candidates` understated that an `--exclude`d file is still read, tokenized
and stored.

### M2 — warm call on m365dotnet (`tools/m2_battery.sh`, RUNS=3)

| id | shape | main median | **branch median** | main peak RSS | **branch peak RSS** |
|---|---|---|---|---|---|
| P1 | file-named | 60.25 s | **17.08 s** | 1615.6 MiB | **1384.7 MiB** |
| P2 | file-named | 54.25 s | **18.45 s** | 1686.2 MiB | **1386.4 MiB** |
| P3 | concept-only | 52.49 s | **19.00 s** | 1801.5 MiB | **1383.3 MiB** |
| P4 | concept-only | 53.52 s | **18.09 s** | 1698.0 MiB | **1383.6 MiB** |
| P5 | mixed | 52.46 s | **18.79 s** | 1851.3 MiB | **1384.2 MiB** |
| P6 | symptom | 39.55 s | **19.27 s** | 1856.8 MiB | **1384.2 MiB** |
| **BATTERY** | | **53.47 s** | **18.50 s** | **1.95 GB** | **1.45 GB** |

**Cold build, reported separately:** m365dotnet 9,353 files in **122.14 s** / 1.47 GB
peak, store 109 MB on disk; neo 307 files in **2.3 s**, store 5 MB. It announces itself
before starting and reports every 250 files, so the minutes are visible rather than
silent.

### M2 targets: 2.9× better, still not met — and the profile says why

Warm profile, m365dotnet, `ru_maxrss` per stage:

```
T imports                      +  1.16s  cum   1.16s  rss=    82MB
T eligibility walk (9348)      +  4.64s  cum   5.80s  rss=    87MB
T content index refresh (warm) +  0.41s  cum   6.22s  rss=   127MB
T content index scores (7266)  +  0.12s  cum   6.33s  rss=   127MB
T project index boost          +  0.12s  cum   6.46s  rss=   140MB
T fact store history boost     +  2.94s  cum   9.40s  rss=  1404MB
```

1. **The cost this goal owns is gone.** Content indexing is **0.53 s of an 18.50 s**
   warm call, down from the ~35–45 s it contributed to main's median.
2. **≤ 500 MB is not reachable from file selection.** 1.26 GB of the 1.4 GB arrives in
   one step — `_history_boost` loading the FactStore (152 MB of JSON carrying 768-dim
   embeddings, ~1.3 GB as Python objects). Goal 1's 1.43 GB baseline was therefore
   never the gatherer's memory; it was the memory system's, measured through the
   gatherer. No change to how files are chosen moves it, and a Goal 6 that "hit
   500 MB" would have done it by deleting a retrieval channel.
3. **The largest remaining wall-clock item is #208's eligibility walk** — 4.6 s for
   9,348 admitted files out of ~345k filesystem entries, CPU-bound and stable across
   repeats (5.88 / 6.91 / 6.63 s in three consecutive in-process calls). Untouched
   here and out of this goal's scope.

Flagging rather than silently narrowing: neither remaining owner is in "move the
per-call BM25 into the on-disk store", and both are named with numbers so the shepherd
can route them. Everything Goal 6 does own is met.

### M3 — freshness cost

```
M3 walk=4.42s  refresh=0.79s  mode=incremental changed=10 added=0 removed=0 indexed=10 total=9353
```

**0.79 s of index work against ≤ 5 s — met.** Quoted with its neighbour so it is not
read as more than it is: walk + refresh is 5.21 s, and the walk is the eligibility scan
every invocation pays regardless of what changed (reading 3 above).

### G1-inv — identical on both arms

| label | distinct | `git check-ignore` excludes | duplicate copies | inside `.worktrees/` | unresolvable |
|---|---|---|---|---|---|
| **BATTERY UNION** | **111** | **0** | **0** | **0** | **0** |

Per-prompt rows in the measurements doc; every cell 0 on both arms — which is the
point, since the index consumes the walker's verdict and cannot widen it.

---

## How Has This Been Tested?

- [x] `tests/test_content_index.py` — 35 new tests in four groups. **Parity**: the same
      corpus scored through both `FileIndex` and `ContentIndex`, compared to floating
      point, over six prompt shapes including repeated terms, camelCase and a path as
      the query, plus rank-order equality. **Freshness**: cold / warm / edit / add /
      delete / newly-gitignored / touch, each asserting both the report and the answer.
      **Degradation**: corrupt store, truncated store, tokenizer bump, schema bump,
      unwritable store, empty repository. **No second walker**: the module is scanned
      for `os.walk`/`glob`.
- [x] Full suite: **2521 passed, 29 skipped, 2 failed** — both failures are
      machine-dependent latency assertions
      (`test_consolidation_performance_with_all_boosts`,
      `test_evaluation_cli_runs_without_provider_keys_or_user_memory`'s
      `performance_within_budget`) and **both fail identically on `origin/main`** in a
      clean worktree, verified before and after this change. This is the class of
      failure CLAUDE.md documents as advisory-not-correctness.
- [x] End-to-end through the real CLI on m365dotnet (9,348 files), aieweb and neo —
      cold, incremental and warm, in both `--dry-run` output modes.

**Test Configuration**: Python 3.13.7 · macOS (darwin 25.5.0) · neo 0.45.0 at
`06c3ae9`

## Checklist

- [x] Code follows the project's style
- [x] Self-review performed
- [x] Commented, particularly the load-bearing decisions (why SQLite, why the cheap
      stamp only decides what to hash, why query-term multiplicity is preserved, why
      corruption is detected at open)
- [x] Docs updated — `CLAUDE.md` retrieval section and a datestamped measurements doc
- [x] No new warnings
- [x] Tests added that prove the change works and that it did not re-rank
- [x] New and existing tests pass (see the two pre-existing failures above)

## Additional Notes

**One stated behaviour difference.** Corpus statistics (N, df, avgdl) are global to the
indexed repository rather than recomputed over each call's filtered subset. The
per-call index recomputed IDF over whatever that call happened to look at, so the same
file could score differently depending on an unrelated flag. Global statistics are
standard IR practice and stable; with no filters — every eval and every default
invocation — the two sets are identical and so are the scores, which the M1 table
above measures rather than assumes.

**`--exclude` no longer prunes a directory subtree.** It still excludes every file
under a named directory (`should_ignore` matches a non-final path component); it just
no longer saves the cost of descending. The names that make pruning matter
(`node_modules`, `.worktrees`, build output) are in the walker's shared default list,
which still prunes them.

**Store location.** `<repo>/.neo/content_index.sqlite3`, beside `index.json`, per the
plan's "on the on-disk store beside the semantic catalog". `.neo` is in the walker's
default ignore list, so Neo never indexes its own index.

Opened as **DRAFT** per the plan's option-3 governance: the shepherd verifies this
evidence and flips it ready; the car-pr-review daemon is the independent gate after.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -719,6 +719,70 @@
   `test_every_chunk_query_compiles` /
   `test_every_edge_query_compiles` prove compilation ONLY, so a new query still
   needs its own behavioural assertion.
+- Persistent content index (`index/content_index.py`, `index/freshness.py`). The
+  BM25 corpus above is **built once and kept on disk**, in the repository's own
+  `.neo/` beside the semantic catalog, not re-derived per call. Rebuilding it per
+  invocation was the entire cost of a Neo call on a large repo: the canonical M2
+  battery on m365dotnet (9,348 eligible files) measured a **53.47 s** median wall
+  and 1.95 GB peak RSS, against 18.50 s / 1.45 GB with the store (#195). Selection
+  is UNCHANGED — byte-identical selected files and rank order on all six battery
+  prompts, and byte-identical `rank_mine_eval` MRR / R@k on all three flagships
+  (neo 0.712, aieweb 0.728, m365dotnet 0.669, 50 cases each).
+  - **SQLite, not the catalog's JSON, and the access shape is the reason.** The
+    catalog is read whole (every embedding participates in every query); a keyword
+    query touches ten terms out of a few hundred thousand, so a parse-whole format
+    would spend the warm budget deserializing postings nothing asked for. Index
+    artifacts only — postings, doc lengths, hashes, a tokenizer/schema signature —
+    never a file body; delivery still reads whole files fresh from disk. Cost is
+    disk: 109 MB for m365dotnet, 5 MB for neo.
+  - **The cheap stamp decides what to HASH, never that a file changed.** size +
+    mtime come free from the walk's existing `stat` (`EligiblePath.mtime_ns`);
+    only a content-hash mismatch counts as a change, so `touch` re-stamps and
+    re-tokenizes nothing and is reported as `touched`, separately from `changed`.
+    `UNRECORDED = -1` disables the cheap path for a store that persists hashes
+    alone — and it is checked EXPLICITLY, because the first cut stamped both the
+    store and the candidate with the sentinel and `-1 == -1` reported every
+    changed file as unchanged.
+  - **Eligibility arrives from the #208 walker and is never recomputed here** —
+    `test_the_module_does_not_walk_the_filesystem` fails on an `os.walk`/`glob` in
+    the module. A file the walker stops admitting (newly gitignored, deleted) has
+    its postings dropped on the same pass, so a stale index cannot answer with an
+    excluded file. Correspondingly the gatherer now walks ONCE, unfiltered, and
+    applies `--exts`/`--include`/`--exclude` to the result: the index is a
+    property of the REPOSITORY, and letting one `--exts py` call prune it would
+    force the next call to rebuild. `--exclude` no longer prunes a directory
+    subtree (it excludes every file under it via `should_ignore`, which matches a
+    non-final component); the names that make pruning matter are in the walker's
+    shared default list.
+  - **Parity is enforced, not asserted.** `K1`/`B`/IDF are imported from
+    `neo.memory.bm25` and the document is the same expression `FileIndex` used, so
+    `TestParity` can score one corpus both ways and compare to floating point.
+    Query-term MULTIPLICITY is preserved — deduping the query before hitting the
+    postings table is a silent ranking change, since the scorer this replaces
+    iterated the token LIST. One stated difference: N/df/avgdl are global to the
+    indexed repository rather than recomputed over each call's filtered subset;
+    with no filters (every eval, every default call) the sets are identical.
+  - Degradations are loud and none is fatal: a **corrupt** store is deleted and
+    rebuilt — detected while OPENING, which matters, because the fallback is
+    permanent and a store that merely failed to open would pin that repository to
+    the full per-call rebuild forever; a **tokenizer/schema** bump wipes and
+    rebuilds (no per-file hash can see it — the files did not change, the
+    tokenizer did); an **unwritable** store still ranks correctly from memory via
+    `FileIndex` and says so. `cold`/`rebuilt` are reported separately though both
+    read everything: only one means something went wrong.
+  - Cold build is bounded and ANNOUNCED before it starts, with progress every 250
+    files — 122 s for m365dotnet's 9,348, 2.3 s for neo's 307. `--dry-run` names
+    which of cold / rebuilt / incremental (N files) / warm / memory happened, in
+    the selected-files block and in the `--json` payload's `content_index` key
+    (`--json` implies `--quiet`, so the stderr note is suppressed on exactly the
+    path a machine consumer reads).
+  - **M2's 500 MB target is NOT reachable from file selection and never was.**
+    Warm profile on m365dotnet: imports 1.2 s, eligibility walk 4.6 s, content
+    index refresh 0.4 s + scores 0.1 s, `_history_boost` 2.9 s **and +1.26 GB**.
+    The RSS is the FactStore (152 MB of JSON with 768-dim embeddings inflating to
+    ~1.3 GB of Python objects) — Goal 1's 1.43 GB baseline was never the
+    gatherer's. What remains of the warm wall-clock is the walker and the memory
+    system, in that order.
 - Context selection (`context_gatherer`): **files are ranked by BM25 over their
   CONTENT** (`neo.file_retrieval`), not by their path. Until 2026-08 they were: the
   scorer took `(rel_path, size, prompt_tokens, git_recent, entry_points)` and the
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -759,17 +759,43 @@
     `TestParity` can score one corpus both ways and compare to floating point.
     Query-term MULTIPLICITY is preserved — deduping the query before hitting the
     postings table is a silent ranking change, since the scorer this replaces
-    iterated the token LIST. One stated difference: N/df/avgdl are global to the
-    indexed repository rather than recomputed over each call's filtered subset;
-    with no filters (every eval, every default call) the sets are identical.
-  - Degradations are loud and none is fatal: a **corrupt** store is deleted and
-    rebuilt — detected while OPENING, which matters, because the fallback is
-    permanent and a store that merely failed to open would pin that repository to
-    the full per-call rebuild forever; a **tokenizer/schema** bump wipes and
-    rebuilds (no per-file hash can see it — the files did not change, the
-    tokenizer did); an **unwritable** store still ranks correctly from memory via
-    `FileIndex` and says so. `cold`/`rebuilt` are reported separately though both
-    read everything: only one means something went wrong.
+    iterated the token LIST. **A filtered call gets filtered statistics**:
+    `scores(prompt, candidates)` takes N, df and avgdl from `candidates`, not
+    from the whole repository. Repo-global stats were the first cut and are the
+    better IR design in the abstract; they are also a re-rank, and a
+    fresh-verifier pass caught it — unflagged runs matched main exactly while
+    `--exts py` changed all 25 selected lines. Now byte-identical under
+    `--exts`, `--exclude` and `--include` as well.
+  - Degradations are loud and none is fatal, and the **except-clause ORDER is the
+    whole mechanism**: `sqlite3.OperationalError` is a SUBCLASS of
+    `DatabaseError`, so `except DatabaseError` written first catches "database is
+    locked" and runs the corruption handler — which `os.unlink`s a perfectly good
+    store. Reproduced: the index was deleted AND the peer's committed transaction
+    went into an unlinked inode and vanished with no error anywhere. Two Neo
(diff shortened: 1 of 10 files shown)


kkeeling:
**Shepherd gate — evidence verified, flipping ready.** CI green. M1 byte-identical on all three flagships. M3 met (10-file refresh 0.79s). M2 battery 53.47s→18.50s / 1.95GB→1.45GB: plan targets (≤5s/≤500MB) NOT met, but the profile attributes the residual outside this goal's scope — content indexing itself is 0.53s of the warm call. M2 stays OPEN in the plan with two named owners: the eligibility walk (4.6s, routing to Goal 7) and FactStore load (2.94s + 1.26GB — separate issue, it was never file selection's cost). Honest flag appreciated; this is exactly what the draft flow is for.

Issue #211, opened 2026-08-13
FactStore history boost loads 1.26GB / 2.9s into every invocation — the memory system, not file selection, owns the M2 memory floor


kkeeling:
From PR #209's warm-call profile on m365dotnet: `_history_boost` loads the FactStore (152MB of JSON carrying 768-dim embeddings) into ~1.3GB of Python objects, adding 2.94s wall and 1.26GB of the 1.4GB peak RSS on every call. Goal 1's 1.43GB M2 baseline was therefore the memory system's cost measured through the gatherer — no file-selection change can move it.

This is OUTSIDE the unified-store plan's scope (docs/unified-store-plan.md) and needs its own decision: lazy/partial FactStore loading, an on-disk vector store, quantized embeddings, or accepting the cost. The plan's M2 memory target (≤500MB) is unreachable without it; per-stage numbers in PR #209.

Filed by the unified-store shepherd so the cost has an owner-shaped ticket rather than hiding inside M2.

PR #212, opened 2026-08-13
perf(walk): persist the eligibility walk so freshness is automatic on every call (#210)

### LINKED ISSUE #210: Eligibility walk costs ~4.6s per warm call on m365dotnet (9.3k admitted / 345k filesystem entries)
From PR #209's warm-call profile: the #208 walker takes 4.6s per invocation on m365dotnet — now the largest wall-clock item in a warm call (content index refresh is 0.41s by comparison). CPU-bound, stable across repeats (5.88/6.91/6.63s consecutive in-process). The M2 plan target (warm ≤5s) is unreachable while every call re-walks 345k entries.

Candidate direction: persist walk results next to the content index and invalidate by directory mtime — i.e., the same freshness treatment PR #209 gave postings, applied to eligibility. Scope-adjacent to plan Goal 7 (auto-freshness); being pulled into its brief.

Plan of record: docs/unified-store-plan.md (M2). Numbers from PR #209's profiled battery.

kkeeling:
## Description

Persists the #208 eligibility walk beside #209's content index, in the repository's own
`.neo/`, so a warm invocation stops re-deriving which files exist and are eligible.
Closes #210. Goal 7 of `docs/unified-store-plan.md`.

Once #209 stopped the BM25 corpus rebuilding per call, the walk was the largest single
item left in a warm Neo call: **4.64 s on m365dotnet** to re-derive that 9,348 files
are eligible, on every invocation, in a repository that had not changed. It is now
**0.16 s**, and the canonical M2 battery went **15.63 s → 8.57 s** median with
byte-identical file selection.

**#209 merged while this was being measured**, so the branch is rebased onto `main`
and the diff is this goal's eight commits alone. The `base` arm in every table below is
`260e56c` (#209's pre-squash head); `git diff 260e56c origin/main` is empty, so it is
content-identical to the squashed `ebd29e0` now on `main` and the comparison stands
unchanged.

### What is cached, and what deliberately is not

- **`index/walk_cache.py`** — per-directory ignore VERDICTS (which subdirectories
  survive, which filenames survive, the exclusion counts), keyed on the directory's
  `mtime_ns` + `ctime_ns`, in `.neo/walk_cache.json`.
- **The design follows from one measurement**, taken before any of this existed. On
  m365dotnet the full walk is 6.85 s, the *same traversal with the per-file ignore test
  removed* is 0.80 s, and `stat`-ing all 9,378 admitted files is 0.10 s. **The walk's
  cost is the pattern matching, not the filesystem** — 6.05 s of 6.85 s is
  `should_ignore`. Caching directory listings would have saved almost nothing; caching
  verdicts saves almost everything.
- **Directory mtime is the right key for one reason**: it moves when an entry is
  created, deleted or renamed, and does NOT move when a file's content changes. An edit
  can change what a file SAYS; it can never change whether it is eligible.
- **File sizes and mtimes are never remembered.** They are read fresh from the `stat`
  the walk owes its callers anyway, because the content index next door uses them as
  ITS freshness stamp. A remembered mtime would make an edited file look unedited —
  one cache's staleness leaking into another's. `TestStampsAreNeverCached` pins it, and
  the M3 evidence below shows an edit landing (`changed=10`) *through* a fully reused
  set of directory listings.
- **A `.gitignore` edit invalidates by CONTENT.** No mtime moves when a pattern file is
  edited, so the signature hashes the effective pattern list (shared defaults + the
  repo's own `.gitignore`/`.ignore`), plus a `MATCHER_VERSION` for the case where the
  patterns are identical and `should_ignore` is not.
- **JSON, not SQLite — the opposite conclusion from the same question #209 asked.**
  Every directory is validated on every call, so the file is read whole every time.
  507 KB and ~10 ms on m365dotnet.
- **`--index` is still an optional warmer.** It goes through the same cache (as does
  `architecture_metrics.compute`, which runs on the outcome path of every real
  invocation), so it warms the walk as well as the catalog — but it is optional,
  because any invocation pays the cold cost once and every later one is warm. A first
  call announces itself before it starts.
- **`--dry-run` reports walk cache hit/miss** — cold / rebuilt / incremental (N of M
  directories) / warm / bypassed, in the selected-files block and in the `--json`
  payload's new `walk_cache` key.

### One defect found in existing behaviour, fixed here

`os.walk(followlinks=False)` had been doing work nobody had named. This traversal
recurses by hand, one directory at a time, so that flag protects only the single read
it is passed: a link to an ancestor was an infinite descent and a link to `/` walked
the machine. The refusal is now explicit, and is not gated by `skip_symlinks` (which is
about whether a symlinked FILE is delivered). The finite case in the same test
(`src/mirror -> docs`) fails in milliseconds rather than hanging.

## Type of Change

- [x] Performance improvement

## Related Issue

Fixes #210

---

## Evidence

Full tables, exact commands and repo HEADs:
**`docs/goal7-walk-cache-measurements-2026-08-13.md`**.
Arms: `main` = `9b0c16d63cfc` (post-#208, pre-#209) · `base` = `260e56c` (#209 head,
content-identical to today's `main` at `ebd29e0`) · `branch` = `0db7695fc003`, rebased
to `a7fa9c7`.

### M2 — warm call on m365dotnet (`tools/m2_battery.sh`, RUNS=3)

| id | shape | main | base (#209) | **branch** |
|---|---|---|---|---|
| P1 | file-named | 58.25 s | 15.91 s | **8.86 s** |
| P2 | file-named | 50.95 s | 15.63 s | **8.69 s** |
| P3 | concept-only | 49.58 s | 15.23 s | **8.07 s** |
| P4 | concept-only | 51.45 s | 15.53 s | **8.37 s** |
| P5 | mixed | 53.14 s | 15.60 s | **8.41 s** |
| P6 | symptom | 52.96 s | 16.17 s | **9.11 s** |
| **BATTERY** | | **52.20 s** | **15.63 s** | **8.57 s** |
| **peak RSS** | | 1893.0 MiB | 1386.8 MiB | 1387.2 MiB |

**Disclosed:** the `main` arm ran at `RUNS=1` (n=1 per prompt) — a 3-run `main` arm is
~16 minutes and this harness caps a foreground command at 10. It agrees with Goal 6's
independently measured 53.47 s for the same arm, which is why it is quoted rather than
dropped, but it is one sample. The machine is a shared laptop whose load varied (28 at
one point); the branch was measured at two shas, 9.47 s and 8.57 s, both published.

### M2 targets: 6× better, and the ≤ 5 s target is met excluding the memory system

Warm profile, m365dotnet, real CLI with timers wrapped around the stages (no profiler,
so the parts sum to a wall-clock a stopwatch would agree with):

```
T imports                                  1.24s
T eligibility walk                         0.16s  (1 call)   <- 4.64s in #209's profile
T content index refresh                    0.11s  (1 call)
T content index scores                     0.12s  (1 call)
T project index boost                      0.10s  (1 call)
T fact store history boost                 2.25s  (1 call)
T fact store retrieve (memory layer)       1.65s  (1 call)
T   of which fastembed model load          2.73s  (2 calls, inside the two above)
T WALL TOTAL                               7.67s   peak rss = 1385 MB
```

1. **The cost this goal owns is gone.** The walk is **0.16 s of a 7.67 s** warm call,
   down from 4.64 s.
2. **File selection is 0.49 s in total** — walk + content index + project index, on a
   9,348-file repository.
3. **Flagged exactly as #209 flagged it, and as the goal brief asks.** Excluding the
   memory system — `_history_boost` (2.25 s) and the engine's own fact retrieval
   (1.65 s), which between them carry both fastembed loads and all 1.26 GB of the RSS
   — the warm call is **3.77 s, under the ≤ 5 s target**. Including them it is 7.67 s.
   Neither is presented as the other. That is issue #211 and out of scope here.

### M3 — freshness cost AND staleness correctness, on the live repository

One script, three change shapes, each timing the full pipeline (walk + index refresh)
and then asserting the answer actually changed. A unique token is written into the
touched files and searched for through the index; the same search after the restore
must find nothing.

```
baseline (settled)     walk= 0.17s [warm, 0/951 re-listed]  refresh= 0.10s [warm]                  TOTAL= 0.27s  files=9348
A: 10 files edited     walk= 0.15s [warm, 0/951 re-listed]  refresh= 0.69s [incremental, chg=10]   TOTAL= 0.84s  files=9348
   -> marker findable in 10 file(s)
A: restored            walk= 0.17s [warm, 0/951 re-listed]  refresh= 0.51s [incremental, chg=10]   TOTAL= 0.68s  files=9348
   -> marker findable in 0 file(s)
B: 10 files added      walk= 0.22s [incremental, 1/951]     refresh= 0.53s [incremental, add=10]   TOTAL= 0.75s  files=9358
   -> marker findable in 10 file(s): src/NeoGoal7Probe0.cs …
B: removed again       walk= 0.25s [incremental, 1/951]     refresh= 0.42s [incremental, rm=10]    TOTAL= 0.67s  files=9348
   -> marker findable in 0 file(s)
C: one file gitignored walk= 5.67s [rebuilt]   files=9347   still eligible: False
C: gitignore restored  walk= 5.30s [rebuilt]   files=9348   back: True

git status --porcelain identical before/after: True
```

That is the staleness evidence the goal asks for — **add a file, gitignore a file,
delete a file, each reflected on the next call** — and the three shapes exercise three
different mechanisms:

- **A: an edit moves no directory's mtime**, so the walk re-lists nothing and the
  index still sees the change, because the file's stamps were read fresh. Had the
  stamps been cached this line would read `warm, 0 changed` and the index would answer
  with text the file no longer holds.
- **B: an add or a delete moves exactly one directory's mtime** — 1 of 951 re-listed.
- **C: a `.gitignore` edit moves no mtime at all**, so the signature discards the cache
  and the walk runs in full once (5.67 s), then warm again.

**Both M3 shapes are met: 0.84 s and 0.75 s against ≤ 5 s.** The live tree was restored
byte-for-byte.

### M1 — retrieval quality: identical, on all three flagships

| repo | arm | cases | MRR | R@1 | R@3 | R@10 | H@10 |
|---|---|---|---|---|---|---|---|
| neo | main | 50 | 0.712 | 0.307 | 0.502 | 0.708 | 0.980 |
| neo | **branch** | 50 | **0.712** | **0.307** | **0.502** | **0.708** | **0.980** |
| aieweb | main | 50 | 0.728 | 0.487 | 0.654 | 0.778 | 0.880 |
| aieweb | **branch** | 50 | **0.728** | **0.487** | **0.654** | **0.778** | **0.880** |
| m365dotnet | main | 8 | 0.906 | 0.583 | 0.729 | 0.958 | 1.000 |
| m365dotnet | **branch** | 8 | **0.906** | **0.583** | **0.729** | **0.958** | **1.000** |

Byte-identical in every cell. **Disclosed:** m365dotnet ran 8 cases, not 50 — its
`main` arm costs ~52 s per case, so 50 cases is ~40 minutes against a 10-minute
foreground cap. Only the two arms of the same 8 cases are comparable with each other,
which is all this table claims; the absolute is not comparable with Goal 6's 50-case
0.669. The branch arms ran at `cc7d98c`, two commits before HEAD — what carries them
to HEAD is the battery re-run at HEAD selecting **byte-identically** to both `cc7d98c`
and `base`.

**Stronger than the MRR table:** all six canonical battery prompts selected the same
files in the same rank order on **all three arms and at both branch shas** — 24
comparisons, no output.

### G1-inv — a cache of the walker's verdicts cannot widen them

| label | distinct | `git check-ignore` excludes | duplicate content | inside `.worktrees/` | unresolvable |
|---|---|---|---|---|---|
| **BATTERY UNION**, main | 111 | 0 | 0 | 0 | 0 |
| **BATTERY UNION**, base | 111 | 0 | 0 | 0 | 0 |
| **BATTERY UNION**, branch | **111** | **0** | **0** | **0** | **0** |

The three union files are byte-identical.

### Cold cost — what a fresh clone pays, once

| repo | files | dirs | uncached walk | first call | second call | cache on disk |
|---|---|---|---|---|---|---|
| neo | 285 | 43 | 0.41 s | 0.36 s | **0.01 s** | 15 KB |
| aieweb | 902 | 221 | 1.05 s | 0.96 s | **0.06 s** | 64 KB |
| m365dotnet | 9,348 | 951 | 6.85 s | 5.3 s | **0.16 s** | 507 KB |

### A fresh non-author verification pass found two real defects

The routing protocol requires a fresh non-author Claude to verify; it did, and it ran a
differential fuzzer (60 random repos × 8 mutation steps × 5 policies), a base-branch A/B
over ten adversarial filesystem shapes plus three real repositories, a five-process
concurrency stress, and twelve source mutations. It found **no input where the cached
walk returned a different file list, order, exclusion count or file stamp** — and two
defects, both fixed in `0db7695`, each with a mutation-verified test:

1. **MEDIUM — `mtime` alone is forgeable.** `touch -r`, `tar -x`, `rsync -a` and every
   snapshot restore write a directory's mtime back to a recorded value, so a restore
   that adds or deletes a file can land on exactly the cached mtime and be reported
   `warm`. Reproduced with two lines of `os.utime`. `st_ctime_ns` moves on any metadata
   change, no userland API restores it, and it arrives in the same `stat` — so the
   defence costs nothing. On Windows `st_ctime` is a creation time and hence constant,
   which makes the extra comparison a no-op there rather than a false invalidation.
   Schema bumped to 2 so existing caches are discarded rather than misread.
2. **LOW/MEDIUM — `cached_walk` on a path that does not exist created it.** The save
   (which `mkdir -p`s `.neo/`) ran even when the walk had read no directory at all, so
   a mistyped `--cwd` silently minted a directory tree — and pointed at an unmounted
   mountpoint it would create a local directory shadowing the mount. Nothing is written
   when nothing was learned, and an empty stored listing map is now rejected on load
   too (believing one made every later call report `incremental` forever).

Also from that pass: secondary walks (`--index`, the arch scan) no longer overwrite the
process-global report the operator's `--dry-run` line describes; and three comments
quoting a *derived* 0.11 s for the directory listing now quote the *measured* 0.80 s
arm, which the measurements doc in the same commit series had contradicted.

## How Has This Been Tested?

- [x] `tests/test_walk_cache.py` — 41 new tests in six groups. **Parity**: the cached
      and uncached walks compared over add / delete / rename / new directory /
      gitignore edit / nested checkout appearing / symlinked directory / policy
      filters. **Live stamps**: an edit is visible through a fully reused listing.
      **Reporting**: every mode is the work actually done. **Racy directories**:
      pre-creating `.neo/` so the guard is what the test measures. **Forged
      timestamps**: `os.utime` restoring an mtime does not hide an add or a delete.
      **Degradation**: corrupt, truncated, empty, malformed-entry, foreign-signature,
      unwritable, missing root, stranded temp files.
- [x] **Seven source mutations, all caught**: constant signature → the gitignore test
      fails; cached file stamps → the stamp test fails; racy-window removed → the racy
      tests fail; descend into symlinked dirs → the symlink test fails; drop the ctime
      key → the forged-timestamp test fails; save a cache of nothing → the missing-root
      test fails; the walk-side `extra_ignores` guard removed → its own test fails. The
      first cut of the racy tests and of the `extra_ignores` test each passed under
      their mutant for an unrelated reason, and both are fixed and re-verified.
- [x] The #208 differential suite against `git check-ignore`
      (`tests/test_eligibility_differential.py`) and the single-source guards
      (`tests/test_eligibility_single_source.py`) pass **untouched** — no change to
      either file.
- [x] Full suite: **2566 passed, 29 skipped, 1 failed**. The failure is
      `test_consolidation_performance_with_all_boosts`, a wall-clock assertion inside a
      correctness test, and it **fails identically on the base branch** (255.6 ms there,
      267.9 ms here, against a 100 ms bound) on this machine. It is the class of failure
      CLAUDE.md documents as advisory-not-correctness.
- [x] End-to-end through the real CLI on m365dotnet (9,348 files), aieweb and neo —
      cold, incremental, warm and gitignore-invalidated, in both `--dry-run` output
      modes.

**Test Configuration**: Python 3.13.7 · macOS (darwin 25.5.0) · neo 0.45.0 at
`0db7695`, rebased onto `main` as `a7fa9c7` with no conflicts and re-verified there

## Checklist

- [x] Code follows the project's style
- [x] Self-review performed, plus an independent fresh-verifier pass
- [x] Commented, particularly the load-bearing decisions (why verdicts and not
      listings, why mtime is the right key, why mtime alone is not enough, why the
      stamps are never cached, why JSON here and SQLite next door)
- [x] Docs updated — `CLAUDE.md` and a datestamped measurements doc
- [x] No new warnings
- [x] Tests added that prove the change works and that it did not re-rank
- [x] New and existing tests pass (see the one pre-existing failure above)

## Additional Notes

**Two clones of one repository do not share a cache.** It lives in each working tree's
own `.neo/`, because it is keyed on that tree's directory stamps.

**A `.gitignore` edit costs one full walk** (5.67 s on m365dotnet), then warm again.
Deriving which directories a pattern edit could have affected is a large amount of
machinery guarding a cost paid when someone edits a `.gitignore`.

**What would still fool it:** a directory whose mtime AND ctime are both restored — not
reachable from normal userland, but a filesystem image edited offline could do it.

Opened as **DRAFT** per the plan's option-3 governance: the shepherd verifies this
evidence and flips it ready; the car-pr-review daemon is the independent gate after.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -728,17 +728,27 @@
   selection on all six prompts and byte-identical `rank_mine_eval` on all three
   flagships.
   - **The cost is the pattern matching, not the filesystem, and that decides the
-    whole design.** Measured before the cache existed: listing the 951 surviving
-    directories 0.11 s, `should_ignore` over their 11,219 entries **6.7 s**,
-    `stat` over all 9,378 admitted files 0.10 s. Caching the syscalls would have
-    saved nothing. So what is stored per directory is the VERDICTS — which
-    subdirectories survive, which filenames survive, and the exclusion counts —
-    and the directory's mtime is what says whether they still hold.
+    whole design.** Measured before the cache existed, on m365dotnet: the full
+    walk **6.85 s**; the same traversal with the per-FILE ignore test removed
+    **0.80 s**; `stat` over all 9,378 admitted files **0.10 s**. So 6.05 s of a
+    6.85 s walk is `should_ignore` (11,219 calls), and caching the syscalls
+    would have saved almost nothing. What is stored per directory is therefore
+    the VERDICTS — which subdirectories survive, which filenames survive, and
+    the exclusion counts — with the directory's stamps saying whether they hold.
   - **Directory mtime is the right key for exactly one reason**: on every POSIX
     filesystem it moves when an entry is created, deleted or renamed inside that
     directory, and does NOT move when the content of a file inside it changes.
     An edit can change what a file SAYS; it can never change whether it is
     eligible.
+  - **mtime alone is not enough, because mtime is forgeable.** `touch -r`,
+    `tar -x`, `rsync -a` and every snapshot restore write a directory's mtime
+    back to a recorded value, so a restore that adds or deletes a file can land
+    on exactly the mtime the cache holds and be reported `warm` — reproduced
+    with two lines of `os.utime`, found by the fresh-verifier pass. `ctime_ns`
+    is stored beside it: the inode change time moves on any metadata change, no
+    API restores it, and it arrives in the same `stat`. On Windows `st_ctime` is
+    a CREATION time and hence constant, which makes the extra comparison a no-op
+    there rather than a false invalidation.
   - **Sizes and mtimes are never remembered.** They come fresh from the `stat`
     the walk owes its callers anyway (0.10 s), because the content index uses
     them as ITS freshness stamp — serving a remembered mtime would make an
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -723,8 +723,8 @@
   The #208 walk is **kept on disk too**, in the same `.neo/`, because once the
   content index stopped rebuilding it became the largest single item in a warm
   call: 4.64 s on m365dotnet to re-derive that 9,348 files are eligible, on every
-  invocation, in a repository that had not changed (#210). Warm walk now 0.2 s;
-  the canonical M2 battery went 15.63 s → **9.47 s** median with byte-identical
+  invocation, in a repository that had not changed (#210). Warm walk now 0.16 s;
+  the canonical M2 battery went 15.63 s → **8.57 s** median with byte-identical
   selection on all six prompts and byte-identical `rank_mine_eval` on all three
   flagships.
   - **The cost is the pattern matching, not the filesystem, and that decides the
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -719,6 +719,80 @@
   `test_every_chunk_query_compiles` /
   `test_every_edge_query_compiles` prove compilation ONLY, so a new query still
   needs its own behavioural assertion.
+- Persistent eligibility walk (`index/walk_cache.py`, `eligibility.DirectoryListing`).
+  The #208 walk is **kept on disk too**, in the same `.neo/`, because once the
+  content index stopped rebuilding it became the largest single item in a warm
+  call: 4.64 s on m365dotnet to re-derive that 9,348 files are eligible, on every
+  invocation, in a repository that had not changed (#210). Warm walk now 0.2 s;
+  the canonical M2 battery went 15.63 s → **9.47 s** median with byte-identical
+  selection on all six prompts and byte-identical `rank_mine_eval` on all three
+  flagships.
+  - **The cost is the pattern matching, not the filesystem, and that decides the
+    whole design.** Measured before the cache existed: listing the 951 surviving
+    directories 0.11 s, `should_ignore` over their 11,219 entries **6.7 s**,
+    `stat` over all 9,378 admitted files 0.10 s. Caching the syscalls would have
+    saved nothing. So what is stored per directory is the VERDICTS — which
+    subdirectories survive, which filenames survive, and the exclusion counts —
+    and the directory's mtime is what says whether they still hold.
+  - **Directory mtime is the right key for exactly one reason**: on every POSIX
+    filesystem it moves when an entry is created, deleted or renamed inside that
+    directory, and does NOT move when the content of a file inside it changes.
+    An edit can change what a file SAYS; it can never change whether it is
+    eligible.
+  - **Sizes and mtimes are never remembered.** They come fresh from the `stat`
+    the walk owes its callers anyway (0.10 s), because the content index uses
+    them as ITS freshness stamp — serving a remembered mtime would make an
+    edited file look unedited and turn one cache's staleness into another's.
+    `TestStampsAreNeverCached` pins it, mutation-verified.
+  - **A `.gitignore` edit invalidates by CONTENT, not by any timestamp.** No
+    directory's mtime moves when a pattern file is edited, so every stored
+    verdict looks current while every one of them may now be wrong. The
+    signature hashes the effective pattern list (shared defaults + the repo's own
+    `.gitignore`/`.ignore`), plus a `MATCHER_VERSION` for the case the patterns
+    are identical and `should_ignore` is not. Cost of an edit: one full walk,
+    then warm again.
+  - **`os.walk(followlinks=False)` had been doing work no one had named.** The
+    traversal now recurses by hand, one directory at a time, so that flag
+    protects only the single read it is passed — a link to an ancestor became an
+    infinite descent and a link to `/` walked the machine. The refusal to
+    descend into a symlinked DIRECTORY is restated explicitly and is not gated by
+    `skip_symlinks`, which is about whether a symlinked FILE is delivered.
+  - **A directory modified within `RACY_WINDOW_NS` (1 s) of being read is not
+    trusted.** Timestamp granularity is not always finer than the interval
+    between two events — HFS+ stamps whole seconds — so a directory read at E
+    with mtime M can be modified afterwards into the same bucket whenever
+    `E - M` is under one tick. Git carries the same guard under the name "racily
+    clean". Its test had to pre-create `.neo/` to mean anything: writing the
+    cache creates that directory, which moves the repository ROOT's mtime, so
+    the call after a first-ever call re-lists the root whatever the guard does —
+    the first cut of that test passed with the guard deleted.
+  - **`extra_ignores` is never served from a cache and never writes one.** A
+    caller's patterns are appended after the repo's own and gitignore is
+    last-match-wins, so a `!negation` there can re-include what a stored verdict
+    excluded — the stored verdict is not a stale answer, it is an answer to a
+    different question. Nothing in Neo takes that path today (`--exclude` is
+    applied after the walk); the guard exists so a future caller gets a correct
+    answer rather than a fast wrong one. Reported as mode `bypassed`.
+  - **JSON, not SQLite — the opposite conclusion from the same question.** Every
+    directory is validated on every call, so the file is read whole every time,
+    which is what JSON is for and what the semantic catalog beside it already
+    uses. 507 KB and ~10 ms for m365dotnet. The neighbouring content index went
+    to SQLite because a query touches ten terms of a few hundred thousand; the
+    access shape decides, not the size.
+  - The verdicts are the IGNORE layer only, so one cache serves every consumer:
+    `--exts` / `match_globs` / `max_file_bytes` / `skip_symlinks` are applied on
+    top, per call. `--index` and `architecture_metrics.compute` (which runs on
+    the outcome-detection path of every real invocation) go through the same
+    cache, so `--index` warms the walk as well as the catalog.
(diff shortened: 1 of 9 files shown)


kkeeling:
**Shepherd gate — evidence verified, flipping ready.** CI green. M2: battery 52.2s (main) → 8.57s; the eligibility walk fell 4.64s → 0.16s (issue #210 closed by this); per-stage profile shows 3.77s excluding the memory layer — the plan's ≤5s wall target is met for everything this plan owns, with the remaining ~4s and the 1.39GB attributed to issue #211 (out of scope, already ticketed). M3 met with staleness correctness shown on the live repo. M1 byte-identical. Honest RUNS=1 main-arm disclosure noted. Daemon review is the independent gate from here.

Issue #213, opened 2026-08-13
`neo --index` builds a catalog of 100% test files on any src/ + tests/ repo


kkeeling:
## What

`ProjectIndex._select_files` ranks candidates **shallowest path first, then alphabetically**:

```python
paths.sort(key=lambda p: (len(p.relative_to(self.repo_root).parts), str(p)))
```

On the conventional `src/<pkg>/…` + `tests/…` layout, every test file is at depth 2 and every source file is at depth 3. So the whole `tests/` tree sorts ahead of the whole `src/` tree, and `--max-files` (default 100) is spent before a single source file is examined.

## Measured, on this repo at `5bbee46`

```
python files by depth: [(2, 105), (3, 47), (4, 38)]
first non-test file in (depth, name) order: rank 102
```

`neo --index` was run and the resulting catalog inspected:

```
indexed files: 99
test files:    99 = 100.0%
sample non-test: []
```

**Zero source files in the catalog.** The build reports success — `[Neo] Built index: 1000 chunks, 777 edges from 100 files` — and the selection report is truthful about the cap (`Indexed 100 of 190 eligible files (capped at --max-files=100)`); nothing says the 100 it kept are all tests.

## Why it matters

The catalog is what the retrieval front door's semantic stage reads. Measured over 50 git-mined cases on this repo (`tools/rank_mine_eval.py`, ground truth = non-test source the commit changed), the pre-#214 `--semantic` lane returned **1224 of 1224 files as test files** and scored **MRR 0.000, R@10 0.000** on every case. Not a ranking weakness — the only files it could return were tests, because they were the only files in the catalog.

This is the same failure shape as #159 (a .NET repo indexed as 83 Python files of a stale worktree, exit 0): a selection rule that is not a ranking, spending a budget in an order nobody chose.

## Not fixed in #214

Goal 8 (one front door) consumes the catalog; it does not build it. #214's front door is robust to this — the same catalog routed through the pipeline's test demotion and keyword channel scores MRR 0.705 instead of 0.000 — but robustness is not a fix, and the semantic stage cannot earn its keep until the catalog holds source.

## Suggested direction

Depth is a proxy for centrality that inverts on the most common Python layout. Options, roughly in order of preference:

1. Apply the gatherer's `is_test_path` demotion to the ranking, so tests fill remaining slots rather than the first ones.
2. Rank by symbol count / file size (the index already computes symbol counts for `_cap_chunks` apportionment).
[...]

PR #214, opened 2026-08-13
feat(retrieval): one front door — named paths, --include, keyword, semantic re-rank

kkeeling:
## Description

Unified-store **Goal 8 (One front door)**. Every invocation now goes through one
retrieval pipeline with one priority order, and the `cli.py` gather fork is gone.

1. **Paths the prompt named** — pinned (`resolve_explicit_paths`)
2. **`--include` patterns** — pinned per standing ruling 1, scan continues
3. **Keyword** — BM25 over the persistent content index (#209)
4. **Semantic** — the embedding catalog, re-ranking and supplementing stage 3
   whenever it exists

Stages 1 and 2 assert presence; 3 and 4 rank what is left. That ordering is the
design: an assertion cannot lose to a score, so a file the caller named is delivered
before the ranking spends a byte.

### `EXPLICIT_PATH_BOOST` becomes a pin

`+10.0` clears every organic signal combined, so a named path did rank first — and
ranking first is a weaker claim than being present. It could still be windowed into a
fragment, and a prompt naming more files than the adaptive limit admits lost the
last-named ones. Stage 1 pins them on the same terms `--include` uses: whole, ahead
of the scan, bypassing `MIN_SCORE_THRESHOLD`, the rank cut and the file cap, with
`--max-bytes` the only thing that may still cut and that cut marked.

The boost **stays on the ranking**. A file can only be pinned if the walk found it,
so the boost covers the candidates the pin pool never held — an `--exts`-narrowed
list, for instance. Deleting it would make the two disagree in exactly the cases
where the pin is unavailable.

### Delivery is one entry per file, read whole from disk

A file over 15,000 characters used to arrive as up to two windows. That cost two
slots of the file cap for one file's worth of coverage, and it is what made
`len(gathered)` and "files" two different numbers (#197). Chunking survives as a
**ranking internal**: it chooses WHICH region of an over-budget file arrives, never
how many entries a file contributes.

`--max-bytes` is apportioned **max-min fair** (`text_budget.apportion`) across the
selection rather than spent greedily in rank order, so a file's SIZE no longer
decides how many other files reach the model — small files arrive whole, large ones
pay. `MIN_FILE_SHARE_BYTES` (512) is the floor below which the ceiling reduces the
file COUNT instead, and `_budget_notes` says by how many and which knob did it.

### What happened to the `cli.py` fork

**It is gone.** `gather_context_semantic`, `mmr_pack_chunks` and
`log_context_metrics` are deleted; `cli.main` calls one function. `--semantic` is a
hint carried on `GatherConfig.semantic` that raises the catalog's weight
(`SEMANTIC_WEIGHT` 1.0 → `SEMANTIC_HINT_WEIGHT` = `CONTENT_WEIGHT` = 3.0) and its
retrieval depth (`SEMANTIC_HINT_DEPTH` = 3×).

Nothing of the lane remains, so there is no "why it must stay" to state. Goal 9 still
owns the wider lane-retirement work; what this PR removes is the fork itself, which
the goal brief scopes here.

Why it had to go rather than shrink: the lane decided whether a *guarantee* applied.
It ignored prompt-named paths entirely, delivered chunks instead of files, and drew
its candidates from the catalog rather than from the walk — so `--exclude` and
`--exts` did not reach it either. A guarantee a second flag switches off is not one.

### One thing deliberately NOT unified

`engine._render_context_files` keeps its own 3,000-character per-file cap for
unpinned files, with its own marker. Collapsing it into the front door's budget would
either grow every prompt by roughly 60% (the renderer's effective ceiling is ~20 ×
3,000 against `--max-bytes` 100,000) or push a relevance-unaware head-cut into the
one place that currently preserves the relevant region. Two budgets, each marked, is
the honest shape until someone measures the merged one.

## Type of Change

## Related Issue

Plan of record: `docs/unified-store-plan.md`, **Goal 8. One front door**.
Surfaces (does not fix) #213.

## Motivation and Context

"Which files does Neo see" had two answers depending on a flag, and the two disagreed
about everything that matters: what is eligible, what is guaranteed, and what a
delivered file even is. The plan's intent is exactly one answer.

---

# Evidence

Full tables, exact commands, repo HEADs and raw harness JSON:
**`docs/goal8-front-door-measurements-2026-08-13.md`** and `evidence/`.

Measured **once, at the end**, against the final base — `origin/main` = `d5adcbc`,
which this branch is already on top of (no rebase was needed). Arms: `main` =
`d5adcbc4e848`, `branch` = `0da5ecb8c845`.

## M1 — retrieval quality, branch vs main, three flagships

`tools/rank_mine_eval.py`, 50 git-mined cases per repo, `--no-git`, 0 failed cases.
Both arms of a repo share `--repo` and differ only by `--tree`; the scorer asserts
the two case lists are equal before scoring.

| repo | arm | n | **MRR** | R@1 | R@3 | **R@10** | H@10 |
|---|---|---|---|---|---|---|---|
| neo | main | 50 | 0.712 | 0.307 | 0.502 | 0.708 | 0.980 |
| neo | **branch** | 50 | **0.712** | 0.307 | 0.502 | **0.708** | 0.980 |
| aieweb | main | 50 | 0.728 | 0.487 | 0.654 | 0.778 | 0.880 |
| aieweb | **branch** | 50 | **0.728** | 0.487 | 0.654 | **0.778** | 0.880 |
| m365dotnet | main | 50 | 0.669 | 0.312 | 0.497 | 0.771 | 0.860 |
| m365dotnet | **branch** | 50 | **0.669** | 0.312 | 0.497 | **0.771** | 0.860 |

**Byte-identical in every cell, on every reported k. No regression on MRR or R@10.**

### Equal metrics, 1,190 more files, none lost

| repo | mean distinct files delivered, main | branch | LOST | GAINED |
|---|---|---|---|---|
| neo | 18.38 | **28.90** | **0** | 526 |
| aieweb | 22.58 | **28.70** | **0** | 306 |
| m365dotnet | 22.14 | **29.30** | **0** | 358 |

In **150 of 150 cases the branch's file set is a strict superset of main's**, and the
top-10 is unchanged in 149 of them. The gain sits below rank 10, which is why the
metrics do not move — R@10 cannot see a file that arrives at rank 24. The cause is
whole-file delivery: a 30-slot budget used to buy 18–22 files' worth of coverage.

The one case whose top-10 moved is stage 1, visible: a query naming `AGENTS.md`,
`CLAUDE.md` and `GEMINI.md` pins the two that exist, so they arrive in path order
rather than score order. Ranks 3 onward identical; neither file is ground truth.

### Concept-only subset, reported separately

A case is **file-named** when its query names a path the repo actually has, tested
with the gatherer's own `extract_explicit_paths` + `matches_explicit_path` against
`git ls-files` — the test that decides which stage runs.

| repo | subset | n | MRR main | MRR branch | R@10 main | R@10 branch |
|---|---|---|---|---|---|---|
| neo | concept | 49 | 0.721 | **0.721** | 0.717 | **0.717** |
| neo | file-named | 1 | 0.250 | **0.250** | 0.250 | **0.250** |
| aieweb | concept | 50 | 0.728 | **0.728** | 0.778 | **0.778** |
| m365dotnet | concept | 50 | 0.669 | **0.669** | 0.771 | **0.771** |

**This instrument cannot show the hoped-for concept win, and the reason is not the
branch.** 149 of 150 mined cases are concept-shaped, so the subset is the corpus and
the split carries no information — and the stage expected to produce the win does not
run: **no flagship has a `.neo/index.json`**, so `_project_index_boost` returns `{}`
on all 300 runs and stage 4 is inert in both arms. Read "no regression" as measured
and "no concept win" as unmeasured-here.

## Stage 4, measured against a repo that has a catalog

Separate experiment, separate repo state — `neo --index` was run first, so this is
not comparable cell-for-cell with the table above. All four arms share that catalog
and the same 50 cases.

| arm | flag | MRR | R@3 | R@10 | mean files | test files delivered |
|---|---|---|---|---|---|---|
| main | — | 0.712 | 0.502 | 0.708 | 18.16 | 8.6% |
| main | `--semantic` | **0.000** | 0.000 | **0.000** | 24.48 | **100.0%** |
| **branch** | — | 0.712 | 0.502 | 0.708 | 28.90 | 10.6% |
| **branch** | `--semantic` | **0.705** | 0.472 | **0.708** | 28.90 | 23.0% |

**The lane scored zero on all 50 cases.** `main --semantic` returned 1,224 files
across 50 cases and **every single one was a test file**. Two compounding causes,
only one of them the lane's:

1. **The catalog contains no source.** `neo --index` selected 99 files; all 99 are
   tests. `ProjectIndex._select_files` ranks shallowest-path-first, and on
   `src/<pkg>/…` + `tests/…` every test is depth 2 while every source file is depth
   3 — 105 Python files at depth 2 here, first non-test at rank 102, `--max-files`
   default 100. Filed as **#213**. It is the index's selection rule, not this goal's,
   and is deliberately not fixed here.
2. **The lane applied none of the pipeline's judgement** — `index.retrieve(k=100)`
   MMR-packed, no test demotion, no keyword channel, no pin. Whatever the catalog
   held, it shipped.

Through the front door the same broken catalog yields 0.705 / R@10 0.708, because it
is one channel of four. **That demonstrates robustness, not quality**, and the doc
says so: the −0.007 MRR under the flag measures the catalog's contents, not the
hint's weight. `SEMANTIC_HINT_WEIGHT` is a defensible default, not a tuned one, and
should be re-measured once #213 lands. `--semantic` is opt-in; the default path is
byte-identical to main.

## M2 — the front door must not re-add cost

`tools/m2_battery.sh`, canonical 6-prompt battery on m365dotnet, 3 timed runs per
prompt after a discarded warm-up, observer disabled, arms sequential. Both arm
launchers were verified to load their own tree before the run (`main` still has
`gather_context_semantic`; `branch` does not).

| | main | branch | delta |
|---|---|---|---|
| battery median wall | 9.10 s | **8.97 s** | −1.4% |
| battery wall range | 7.62–10.44 s | 7.67–10.51 s | — |
| battery peak `ru_maxrss` | 1387.2 MiB | **1390.5 MiB** | +0.24% |

Both deltas are inside the noise floor — main's own runs span 7.62 s to 10.44 s, so a
0.13 s median difference is not a signal. Within noise of main's warm profile, which
is the requirement.

### Selection on the same battery

| | main | branch |
|---|---|---|
| context entries emitted | 180 | 151 |
| sum of per-prompt distinct files | 135 | 151 |
| **within-prompt repeat entries** | **45** | **0** |
| battery union, distinct files | 111 | **125** |

`entries == distinct files` on the branch, all six prompts. #197's divergence is now
impossible rather than merely reported.

## G1-inv

| | main | branch |
|---|---|---|
| union size | 111 | 125 |
| `git check-ignore` hits | **0** | **0** |
| byte-identical duplicate copies | **0** | **0** |

Duplicates tested on CONTENT. Three basenames repeat in the branch union
(`CURRENT.md` ×8, `INDEX.md` ×2, `AUTH_AND_SECURITY.md` ×2) and each is a genuinely
different file; counting shared basenames would report a break on a repo that has
none.

## How Has This Been Tested?

- [x] **Guard-invariant battery** — `pytest -m invariants`: **100 passed, 1 skipped**.
      Gitignored, duplicates, the `--include` guarantee, the prompt-named guarantee
      and the truncation markers, each per language (C#, TypeScript, Python).
- [x] **`tests/test_front_door.py` — 26 new tests.** Stage 1 as a pin, stage order,
      one entry per file, the marker measured against the whole file, max-min fair
      apportionment, and `--semantic` as a weight/depth hint with no lane behind it.
      Every guarantee assertion is paired with a control that fails without it —
      since delivery became whole-file, "nothing was windowed" is the DEFAULT, so
      both arms of the pin tests run at the SAME ceiling and differ only in the flag.
- [x] **Full suite: 2,675 passed, 29 skipped, 1 failed.** The failure is
      `test_consolidation_performance_with_all_boosts`, a `_merge_cluster` wall-clock
      assertion with no relation to context gathering, and it is **reproduced on
      unmodified `origin/main` in a separate worktree** (176.3 ms there, 150.6 ms
      here, against a 100 ms bound) — a documented property of this machine, green in
      CI. It is the same row #207's table carries.

**Test Configuration**: Python 3.13.7 · macOS 26.5.2 (Apple M4 Max) · neo 0.45.0 ·
base `origin/main` = `d5adcbc`

### Four existing tests changed because their premise did

- The G3-inv specimen moved from the prompt-named target to a new oversized file the
  prompt does NOT name (`bulk_rel`). A pin is delivered whole and the renderer does
  not cut it, so the target can no longer prove "every cut is marked" — left alone,
  that assertion would have gone quietly vacuous rather than red.
- `_limit_for` in the invariant battery now knows about pins, which
  `_render_context_files` has always done.
- `test_an_included_file_is_not_windowed` → `_is_not_excerpted`: both arms at one
  ceiling, because a generous-ceiling control proves nothing now.
- The byte-cap warning test follows the mechanism: the ceiling reduces the file COUNT
  and says by how many, instead of stopping the fill loop silently.

## Checklist

## Additional Notes

Opened as **DRAFT** per `docs/unified-store-plan.md`'s option-3 governance: the
shepherd verifies the evidence above against the plan and flips it ready; the
car-pr-review daemon is the independent gate after that.

`tools/rank_mine_eval.py` gained two things it needed to answer this DoD: `--semantic`
(the hint is a CLI flag, so measuring it means passing it) and a `per_case` block in
`--json`, so a SUBSET can be scored without re-running 50 subprocesses. Re-mining with
a filter would draw a second sample and the two arms would no longer be scored on
identical cases.


kkeeling:
**Shepherd gate — evidence verified, flipping ready.** CI green (an earlier red run was fixed and re-run by the goal before I gated — the conveyor working as intended). M1 byte-identical on all flagships including the concept-only subset; M2 within noise (8.97s vs 9.10s); the gather fork is deleted outright — one pipeline, --semantic now a weight hint. Priority order matches the plan: named paths → --include (ruling 1) → keyword → semantic re-rank. Daemon review is the independent gate from here.

PR #215, opened 2026-08-13
fix(front-door): pins cannot starve the scan; each budget cap owns its own count

kkeeling:
## Description

Follow-up to #214 (unified-store Goal 8). Two defects that #214 introduced and shipped, both found by a fresh non-author verifier reading **#214's own committed evidence** after it merged. Neither is a new feature; both restore properties #214 claimed and broke.

### 1. The pin block could spend the whole byte ceiling, and the scan got nothing

Standing ruling 1 is *"guarantee the named files … AND keep scanning for anything else useful"*. Funding pins to the last byte satisfies the first clause by deleting the second.

Reproduced on `main` at `5c0ce5d`, using the canonical M2 battery's own P1 prompt — it names `src/Parslee.M365.Api/Program.cs` (442,867 bytes):

| | pre-#214 (`d5adcbc`) | **#214 as merged (`5c0ce5d`)** | this PR |
|---|---|---|---|
| entries / distinct files | 30 / 22 | **1 / 1** | 30 / 30 |
| context bytes | 240,525 | 299,959 | 284,406 |

The pin took 299,959 of the 300,000-byte default, `remaining_bytes` fell to 41, `affordable` to 0, and **the entire context was one file**. Mentioning a large file in your prompt deleted the rest of your context.

`PIN_BUDGET_SHARE` (0.5) holds the pin block to half of `--max-bytes` **while the scan still has candidates of its own**. The held-back cut is marked inline and announced — which the ruling explicitly permits ("whole, or with an explicit marker"). With nothing else eligible the reserve would fund nothing, so it does not apply and the pin arrives whole; `test_pins_take_the_whole_ceiling_when_nothing_else_is_eligible` pins that.

Half, and symmetric, because neither of "the files you named" and "everything else useful" has a claim to more of the other's budget. It binds on exactly one of the six battery prompts.

### 2. Both budget caps could bind and only the file cap was named

`file_cap_rejected` was derived from the full candidate list *before* the byte cap had cut anything, so whenever both bound the file cap was reported and the byte cap was never mentioned. On the same P1 run, merged main prints:

```
Warning: pinned files filled the file budget (--max-files=30); the scan contributed nothing
```

— with **29 of 30 file slots free** and the byte ceiling holding the count at one. Raising the knob the run named provably changes nothing.

Each cap is now charged for what IT removed, in application order (`dropped_by_file_cap` counted before the byte cap cuts, `dropped_by_byte_cap` after), and **both are reported when both bind** with their own counts. In the scan-delivered-nothing branch the byte cap is tested first, because it is applied second and therefore holds the margin.

This is the repo's own `never blame a cap for an absence it did not cause` rule, broken inside the goal whose subject is selection truthfulness.

### 3. Three of #214's new tests did not bite

Verified by mutation, not by reading:

- `test_a_named_path_survives_a_file_cap_of_one` passed with **stage 1 deleted** — the fixture named ONE file, and `EXPLICIT_PATH_BOOST` alone still delivers one file into one slot. It now names THREE paths against a one-file cap, which boosting cannot satisfy.
- `test_pins_come_before_the_scan` passed **vacuously** with zero pins: `[] == list(range(0))`. It now asserts the pin list is non-empty and that the scan also contributed.
- `test_every_ranked_file_arrives_despite_one_large_one` passed with `apportion` replaced by a greedy fill, because a zero-allowance file still "arrives" as a marker-only entry. It now asserts the small files arrive **byte-identical to disk**.

Both fixtures gain a second large file: with only one, no single ceiling can separate pinned from unpinned, because any ceiling that leaves the pin block room also leaves the scan's fair share room.

## Type of Change

- [x] Documentation update

## Related Issue

Follows up #214. Plan of record: `docs/unified-store-plan.md`, Goal 8.

## Motivation and Context

#214 was measured, reviewed and merged, and still shipped a case where naming a file in your prompt reduced your context to that one file. It was visible the whole time in `evidence/m2_branch/P1_run1.files` — one line, `entries=1` — and the measurements doc reported P1's wall-clock and RSS and nothing else. The doc is corrected in place rather than rewritten, including the confound this puts on the M2 median: the branch's P1 did ~1/30th of main's delivery work and was still the slowest branch prompt, so that median is not like-for-like.

## How Has This Been Tested?

- [x] **Both defects reproduced on merged `main` and shown fixed**, side by side, via `neo --dry-run` on the P1 prompt against `m365dotnet` — see the table above. Selection counts, not timings, so the result is contention-free and deterministic.
- [x] **`tests/test_front_door.py` — 9 new tests** in two classes (`TestThePinBlockCannotStarveTheScan`, `TestEachCapIsChargedForWhatItRemoved`), including `test_raising_the_named_knob_actually_moves_the_number`, which is the property a mis-attributed warning breaks.
- [x] **Mutation-verified, both fixes.** `PIN_BUDGET_SHARE = 1.0` → 3 failures. Reinstating the old cap attribution → 2 failures. A fix whose test stays green when the fix is reverted is what produced defect 3.
- [x] `pytest tests/test_front_door.py tests/test_include_guarantee.py` → **99 passed**
- [x] `pytest -m invariants` → **100 passed, 1 skipped**
- [x] `ruff check src/ tests/ tools/` clean

**Test Configuration**: Python 3.13.7 · macOS 26.5.2 (Apple M4 Max) · base `origin/main` = `5c0ce5d`

### M1 is not re-run, deliberately

`PIN_BUDGET_SHARE` binds only when the pin block would exceed half the ceiling, and exactly 1 of the 150 git-mined M1 cases names a path at all (it names three small markdown files). The unified-store **Goal 9** session independently re-ran M1 at the post-#214 base and reproduced #214's branch arm byte-for-byte on all three flagships — neo 0.712 / 0.708, aieweb 0.728 / 0.778, m365dotnet 0.669 / 0.771, 0 failed cases. Different session, different base, same numbers; cited rather than re-spent, and credited to that run.

M2 wall-clock is **not** re-measured here: another goal's battery had the laptop, and a contended timing number would be worse than none. This PR's claim is about selection, which is deterministic.

## Checklist

## Additional Notes

Opened as **DRAFT** per `docs/unified-store-plan.md`'s option-3 governance for runtime-touching changes: the shepherd verifies the evidence and flips it ready; the car-pr-review daemon is the gate after that.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1003,6 +1003,27 @@
   on the same terms `--include` uses. The boost STAYS on the ranking: a file can
   only be pinned if the walk found it, so the boost covers the candidates the
   pin pool never held (an `--exts`-narrowed list).
+  **The pin block cannot spend the whole ceiling** (`PIN_BUDGET_SHARE` = 0.5).
+  Ruling 1 is "the named files AND keep scanning", and funding pins to the last
+  byte satisfies the first clause by deleting the second. Measured on the M2
+  battery after #214 merged: the prompt naming
+  `src/Parslee.M365.Api/Program.cs` (442,867 bytes) pinned it, spent 299,959 of
+  the 300,000-byte default, and delivered **one file** — the whole context was
+  that file — where pre-#214 main delivered 22 and the fix delivers 30. The
+  reserve binds only when pins would take more than half; with nothing else
+  eligible it would fund nothing, so it does not apply and the pin arrives
+  whole. The held-back cut is marked and announced, which the ruling permits
+  ("whole, or with an explicit marker").
+  **Each budget cap is charged for what IT removed.** `dropped_by_file_cap` is
+  counted before the byte cap cuts and `dropped_by_byte_cap` after, and both are
+  reported when both bind. Deriving the file cap's verdict from the full
+  candidate list made the run print `pinned files filled the file budget
+  (--max-files=30)` with 29 of 30 slots free and the BYTE ceiling holding the
+  count at one — raising the named knob provably changed nothing, which is this
+  repo's own rule about never blaming a cap for an absence it did not cause,
+  broken inside the goal whose subject is selection truthfulness. In the
+  scan-delivered-nothing branch the byte cap is tested FIRST, because it is
+  applied second and therefore holds the margin.
   **Delivery is one entry per file, read whole from disk.** Chunking survives
   only as a RANKING internal — it chooses WHICH region of an over-budget file
   arrives, never how many entries a file contributes. `--max-bytes` is
--- a/docs/goal8-front-door-measurements-2026-08-13.md
+++ b/docs/goal8-front-door-measurements-2026-08-13.md
@@ -198,6 +198,41 @@ P1 and P5 are slower and P2, P3, P4, P6 faster; every one of those differences i
 smaller than the spread within a single prompt's own three runs. Read the battery
 median, not a row.
 
+### Correction (2026-08-13, post-merge): P1 delivered ONE file, and this doc did not say so
+
+The table above reports P1's wall-clock and RSS and nothing else, and the
+selection table below reports only aggregates. Both concealed the single largest
+behavioural change on the battery, which a fresh non-author verifier found in
+this branch's own committed artifact:
+
+| | pre-#214 main | #214 as merged | with `PIN_BUDGET_SHARE` |
+|---|---|---|---|
+| P1 entries / distinct files | 30 / 22 | **1 / 1** | 30 / 30 |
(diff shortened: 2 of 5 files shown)


kkeeling:
**Shepherd gate — evidence verified, flipping ready.** CI green. This restores standing ruling 1's second clause (the scan survives pins) with a reproduced counterexample from the canonical battery's own P1 prompt — a defect every gate including mine missed on #214; credit to the fresh-verifier pass. PIN_BUDGET_SHARE symmetry rationale is sound; the nothing-else-eligible carve-out is pinned by test. Flipping first; #216 follows after this lands.

